PR #9, opened 2018-02-08
Fix indentation for subsections 2.1.3 - 2.1.5

jordan-wright:
The sections 2.1.1 and 2.1.2 are nested one layer below the section header 2.1, which makes sense. However, this indentation isn't consistent with sections 2.1.3 - 2.1.5. This change simply updates the indentation to keep all the role sections consistent.


vladimir-v-diaz:
👍 LGTM.

Issue #13, opened 2018-02-27
Extraneous word ("developer") in spec section 2.1.2


awwad:
Section 2.1.2 of the TUF spec refers to "Delegated developer roles" suddenly. I'm guessing this was missed in a prior change and should just read "delegated roles". You can see this in the second paragraph below.

> In addition, the targets role can delegate full or partial trust to other roles. Delegating trust means that the targets role indicates another role (that is, another set of keys and the threshold required for trust) is trusted to sign target file metadata. Partial trust delegation is when the delegated role is only trusted for some of the target files that the delegating role is trusted for.
>
> Delegated developer roles can further delegate trust to other delegated roles. This provides for multiple levels of trust delegation where each role can delegate full or partial trust for the target files they are trusted for. The delegating role in these cases is still trusted. That is, a role does not become untrusted when it has delegated trust.

PR #14, opened 2018-02-27
Clarify wording in 2.1.2 Targets role re. revocation

awwad:
Make it clear that delegation is whitelisting, and revocation is
removal from the whitelist. Prior text implied a blacklist
("`signing new metadata that indicates the delegated role is no
longer trusted`")

Also, strip an extraneous word ("developer" in "Delegated developer
roles"), fixing #13.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,6 +1,6 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **9 February 2018**
+Last modified: **28 February 2018**
 
 Version: **1.0 (Draft)**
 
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -298,14 +298,14 @@ Version: **1.0 (Draft)**
...
-      Delegated developer roles can further delegate trust to other delegated
+      Delegated roles can further delegate trust to other delegated
   roles.  This provides for multiple levels of trust delegation where each
   role can delegate full or partial trust for the target files they are
   trusted for.  The delegating role in these cases is still trusted.  That is,
   a role does not become untrusted when it has delegated trust.
 
       Delegated trust can be revoked at any time by the delegating role signing
-  new metadata that indicates the delegated role is no longer trusted.
+  new metadata that no longer includes that delegation.
```
vladimir-v-diaz: Something feels off with the modified sentence.  Do you find the following revision easier to understand?

> The Targets role, or a delegated role, can revoke trust by signing new metadata that no longer includes that delegation.

vladimir-v-diaz:
This pull request should also update the "last modified" field.

tuf-spec.md
awwad: You're right: it's a little awkward. The replacement is ambiguous at "or a delegated role", unfortunately. (It's not clear if "delegated role" is the delegatee or just any non-Targets role --
-- which is what you mean.) Perhaps "or any delegating role", so that it doesn't sound like it's the delegatee.

Or, how about:
`Any delegation can be revoked at any time: the delegating role need only sign new metadata that no longer contains that delegation.`

Maybe we can add this to that if you think it is needed:
`The current state of trust is always represented by the most recent version of metadata.`

vladimir-v-diaz: "Any delegation can be revoked at any time: the delegating role need only sign new metadata that no longer contains that delegation."

That's clear to me.

awwad: K, I'll amend the commit to that instead, then (and update last modified).

PR #22, opened 2018-06-19
Fix subsections of detailed workflow: Iissue #16

### LINKED ISSUE #16: Spec numbering and bulletting is a bit confusing
See, for example, the section 5.0 and 5.1 headings [here](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows). 5.1 is on the same outline level as 1.1 below it, and as a result, it's a bit hard to specify, say, item 4.5.1 in section 5 (or I guess it's in section 5.1, for that matter?).

vladimir-v-diaz:
Close #16.

This pull request fixes the indentation and numbering of the detailed workflow.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -994,7 +994,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 ## **5. Detailed Workflows**
 
-    **The client application**
+    ### **The client application**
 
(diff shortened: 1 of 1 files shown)


vladimir-v-diaz:
Thanks for the review!

PR #23, opened 2018-06-19
Update section 4.5 of the detailed client workflow.  Fix issue #17

### LINKED ISSUE #17: Clarify consequence of a delegated-to role not being found during target validation
## Summary
The spec is not adequately clear on what a client should do when traversing a list of delegations if it cannot obtain and validate a role specified in that list. As a consequence, different clients may have different delegation order semantics -- intentionally or unintentionally.

## Background
Near the bottom of section 4.5 [in the spec](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats) is an explanation of delegation prioritization.

## Details
Scenario:
Suppose some role "T" makes a non-terminating delegation of target files matching `some_path/*` to some roles "R1" and "R2", in that order.
_(Note that the question is moot for a terminating delegation. "Terminating" is defined in the spec, but in short: if T makes a terminating delegation of namespace some_path/\* to R1, then if the search for a target that matches some_path/\* reaches the delegation to R1, it will not backtrack to proceed to R2 or further delegations in T's list of delegations. In other words, by making a terminating delegation of some_path/\* to R1, T is excluding from its later delegations the right to specify a target that matches some_path/\*)_

**Early delegation confers early right to specify a target. It is clear that if R1 specifies target info and a client finds a matching target file, that this should be considered the valid target file, and any target info R2 provides for this target should be disregarded.** But what if a client can't find or validate an R1 role file? Should the client proceed to R2?

## Solution
**The behavior we think makes the most sense, to provide for a consistent meaning for delegations and prevent MITM attacks that exclude particular role files resulting in different targets being deemed valid, has the client halt its search for target file info for the given target if it cannot obtain or validate R1, and not proceed to R2 or backtrack, since R1 has the right to specify the target info for that target.** This also is the behavior of the reference implementation.

The spec should indicate that this is the expected behavior, in [spec section 5.1 (Detailed client workflows)](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows).

If this is not made rigid, it should at least be made clear that delegation order semantics will vary based on the behavior chosen. (People should not stumble into a behavior.)

vladimir-v-diaz:
Close #17.

This pull request updates the text in section 4.5 of the detailed workflow to explain that requested metadata that cannot be downloaded nor validated should cause the search to halt.

Signed-off-by: Vladimir Diaz \<vladimir.v.diaz@gmail.com>

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1135,7 +1135,9 @@ non-volatile storage as FILENAME.EXT.
   the new targets metadata file becomes the trusted targets metadata file.
 
   * **4.5**. **Perform a preorder depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.**
+  desired target, beginning with the top-level targets role.**  If any metadata
+  requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor validated, end
+  the search and report that the target cannot be found.
 
     * **4.5.1**. If this role has been visited before, then skip this role (so
     that cycles in the delegation graph are avoided).  Otherwise, if an
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1135,9 +1135,9 @@ non-volatile storage as FILENAME.EXT.
   the new targets metadata file becomes the trusted targets metadata file.
 
   * **4.5**. **Perform a preorder depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.**  If any metadata
-  requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor validated, end
-  the search and report that the target cannot be found.
+  desired target, beginning with the top-level targets role.**  Note: If
+  metadata requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor
+  validated, end the search and report that the target cannot be found.
 
     * **4.5.1**. If this role has been visited before, then skip this role (so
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1135,7 +1135,9 @@ non-volatile storage as FILENAME.EXT.
   the new targets metadata file becomes the trusted targets metadata file.
 
   * **4.5**. **Perform a preorder depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.**
+  desired target, beginning with the top-level targets role.**  Note: If
+  metadata requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor
+  validated, end the search and report that the target cannot be found.
```
awwad: This technically covers #17, but can we add "any" in this line?

Note: If **any** metadata requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor
validated, end the search and report that the target cannot be found.

I can imagine an implementer seeing the behavior of halting in a sequence of delegations if the first cannot be retrieved as unintended, so emphasis seems called for. Perhaps that's too paranoid.

vladimir-v-diaz: Okay, I revised the text in https://github.com/theupdateframework/specification/pull/24/commits/f57722c05ed1fba04862fefa5225048862e0c168.

PR #24, opened 2018-06-20
Report failure for update steps that require it.  Fix issue #21

### LINKED ISSUE #21: Spec's detailed workflow should include steps to follow after metadata validation fails
The [detailed workflow section](https://github.com/theupdateframework/specification/blob/409739f2b8601e28d9330eeabeb454d9ef188e7d/tuf-spec.md#5-detailed-workflows) of the spec should clarify what clients should do after a failed metadata check.  For instance, subsection **1.3 check signatures** requires the client to verify that the metadata is properly signed, but what should happen if the metadata fails the check?  The given metadata should definitely be discarded, but what about the client's current version of metadata?

vladimir-v-diaz:
Close #21.

This pull request revises some of the steps in the detailed workflow to cover the case where a failure should be reported.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1077,7 +1077,8 @@ used to download the timestamp metadata file is of the fixed form FILENAME.EXT
   * **2.3**. **Check for a freeze attack.** The latest known time should be
   lower than the expiration timestamp in the new timestamp metadata file.  If
   so, the new timestamp metadata file becomes the trusted timestamp metadata
-  file.
+  file.  If the new timestamp metadata file has expired, discard it and report
+  the potential freeze attack.
 
 **3**. **Download snapshot metadata file**, up to the number of bytes specified
 in the timestamp metadata file.  If consistent snapshots are not used (see
@@ -1090,10 +1091,13 @@ non-volatile storage as FILENAME.EXT.
 
   * **3.1**. **Check against timestamp metadata.** The hashes and version number
   of the new snapshot metadata file MUST match the hashes and version number
-  listed in timestamp metadata.
+  listed in timestamp metadata.  If hashes and version do not match, discard
+  the new snapshot metadata and report the failure.
 
-  * **3.2**. **Check signatures.** The snapshot metadata file MUST have been
-  signed by a threshold of keys specified in the trusted root metadata file.
+  * **3.2**. **Check signatures.** The new snapshot metadata file MUST have
+  been signed by a threshold of keys specified in the trusted root metadata
+  file.  If the new snapshot metadata file is not signed as required, discard
+  it and report the signature failure.
 
   * **3.3**. **Check for a rollback attack.**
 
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -996,6 +996,10 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 ### **The client application**
 
+  Note: If a step in the following workflow does not succeed (e.g., a failure
+  is reported because a new metadata file was not signed), the client should
+  not be left in an unrecoverable state.
+
```
awwad: Two points:

1. `client should not be left in an unrecoverable state` seems less specific than it could be. Consider instead: `the client must be able to attempt update again`?

2. Steps below say "report the signature failure" or "report ... attack", but reporting doesn't necessarily mean aborting the update cycle. Would we want implementers to continue with existing unexpired versions of metadata? If timestamp fails to validate, or a new version of root fails to validate, should we continue the update cycle using the trusted unexpired versions? I think no. So consider "If any metadata fails to validate, abort the update cycle".

tuf-spec.md
```diff
@@ -1020,13 +1024,18 @@ repo](https://github.com/theupdateframework/specification/issues).
   * **1.3. Check signatures.** Version N+1 of the root metadata file MUST have
   been signed by: (1) a threshold of keys specified in the trusted root
   metadata file (version N), and (2) a threshold of keys specified in the new
-  root metadata file being validated (version N+1).
+  root metadata file being validated (version N+1).  If version N+1 is not
+  signed as required, discard it and report the signature failure.  On the
+  next update cycle, begin at step 0 and version N of the root metadata
+  file.
```
awwad: 👍 

Should we also say "abort this update cycle"?

tuf-spec.md
```diff
@@ -1020,13 +1024,18 @@ repo](https://github.com/theupdateframework/specification/issues).
...
 
   * **1.4. Check for a rollback attack.** The version number of the trusted
   root metadata file (version N) must be less than or equal to the version
   number of the new root metadata file (version N+1). Effectively, this means
   checking that the version number signed in the new root metadata file is
-  indeed N+1.
+  indeed N+1.  If the new root metadata file is less than the trusted metadat
```
awwad: IMO need to write "version of". Also need to correct typo.

> If the **version of** the new root metadata file is less than the **version of** the trusted **metadata**

tuf-spec.md
```diff
@@ -1057,16 +1069,20 @@ used to download the timestamp metadata file is of the fixed form FILENAME.EXT
 
   * **2.1**. **Check signatures.** The new timestamp metadata file must have
   been signed by a threshold of keys specified in the trusted root metadata
-  file.
+  file.  If the new timestamp metadata file is not properly signed, discard it
+  and report the signature failure.
```
awwad: Is it clear enough in this addition that the update cycle should be aborted?

Comment holds for the modifications above and below, too, though that can be clarified once at the top instead (see first review comment).

vladimir-v-diaz:
I revised many of the sections according to your feedback.  Thanks for the review!

PR #35, opened 2019-03-04
Slightly clarifying provenance / early history...

JustinCappos:
It's odd to fix this so late in the project's lifecycle, but the spec should probably be a bit more accurate in describing the history of TUF and Thandy.  Jake and Roger visited UW (while Justin S. and Justin C. were there) and the brainstorming discussions we had there laid the groundwork for Thandy, which Nick, Jake, Roger, and possibly others I am unaware of designed.  The Thandy authors then reached out to Justin S. and Justin C. to examine Thandy.  The creation of TUF (by Justin S. and Justin C.) came from looking at issues with Thandy where we thought security could be improved (e.g., the lack of a snapshot role).  We also tried to build TUF as a library so that others did not need to do a design like Thandy in order to have a secure updater.

Note, this is my recollection of events and memory is not perfect.  If someone else has a different recollection about any of my "facts" above, please let me know and I can adjust or annotate what is written.

### DIFF
--- a/tuf-spec.0.9.txt
+++ b/tuf-spec.0.9.txt
@@ -39,17 +39,16 @@ Version 0.9
    (http://www.geni.net/)
    (http://www.nsf.gov/)
 
-   TUF's reference implementation is based heavily on Thandy, the application
-   updater for Tor (http://www.torproject.org/). Its design and this spec are
-   also largely based on Thandy's, with many parts being directly borrowed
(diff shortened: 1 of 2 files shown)


awwad:
@JustinCappos  Since I edited it a wee bit, you might want to take a quick peek, but it LGTM now and I'm merging it.

PR #37, opened 2019-03-19
0.9 timestamp example was missing "version"

erickt:
The 0.9 tuf spec in section 4.6 mentions that "version" is included in the signed portion of timestamp.json, but was missing from the example file. This adds it.

Feel free to reject this patch if you want to leave 0.9 as a historical document.


JustinCappos:
Good catch.  Thanks for fixing.

PR #39, opened 2019-03-28
Recommend against leading path separators

erickt:
In https://github.com/theupdateframework/tuf/pull/679, the python tuf project forbade leading path separators due to surprising behavior joining target and delegated paths, since Python in particular has the following surprising behavior:

```
>>> os.path.join("/foo", "/bar")
'/bar'
```

This PR updates the spec to recommend against using leading path separators in targets.json, and removes their use from the examples. Does this change look acceptable, and is "should not" be the correct phrase, or can this be the more strict "must not"?

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -766,7 +766,9 @@ repo](https://github.com/theupdateframework/specification/issues).
        }
 
    Each key of the TARGETS object is a TARGETPATH.  A TARGETPATH is a path to
-   a file that is relative to a mirror's base URL of targets.
+   a file that is relative to a mirror's base URL of targets. It should not
+   have a leading path separator to avoid surprising behavior when constructing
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
Wish I could merge, but thanks Erick!

erickt:
Hello, just checking in again. Do you need me to make any changes, or do you think it's not worth merging this in?

erickt:
(hit comment a second too early). I could change the language to use "RECOMMENDED not to use leading slashes" if you want to allow implementations to still use leading slashes if they want to?

erickt:
so fast! thanks @trishankatdatadog!

trishankatdatadog:
I think we should be OK. Haven't seen anyone else complain about this besides myself.

PR #40, opened 2019-04-02
Remove root, add delegation hashes to the snapshot metadata

### LINKED ISSUE #31: Snapshot.json example is missing hash(es) and size for the root role
In section [4.6 of the spec](https://github.com/theupdateframework/specification/blob/99dbf9af48c9d8c3cc78ce1b8d185df10aa8c503/tuf-spec.md), the description of the snapshot.json file states:

> The snapshot.json file is signed by the snapshot role. It lists the version numbers of all metadata on the repository, excluding timestamp.json and mirrors.json. For the root role, the hash(es), size, and version number are listed.

However, the example does not include the hashes or size for the root.json file:

```
 {
 "signatures": [
  {
   "keyid": "fce9cf1cc86b0945d6a042f334026f31ed8e4ee1510218f198e8d3f191d15309",
   "sig": "f7f03b13e3f4a78a23561419fc0dd741a637e49ee671251be9f8f3fceedfc112e4
           4ee3aaff2278fad9164ab039118d4dc53f22f94900dae9a147aa4d35dcfc0f"
  }
 ],
 "signed": {
  "_type": "snapshot",
  "spec_version": "1",
  "expires": "2030-01-01T00:00:00Z",
  "meta": {
   "root.json": {
    "version": 1
   },
   "targets.json": {
    "version": 1
   },
   "project.json": {
    "version": 1
    },
   }
  "version": 1
  },
 }
```

Could they be added?

Also, I've seen mentioned in a [few places that TAP 5](https://github.com/theupdateframework/tuf/issues/451) removes the need to actually  include the root in the snapshot file, but I'm not sure if TAP 5 is integrated into the spec, or is just planned to be incorporated in some future version of TUF.

erickt:
According to @JustinCappos in #31, the root.json is no longer needed in the snapshot metadata, because the workflow will already have updated the root metadata before the snapshot metadata is fetched.

In addition, section 5.6 of the Mercury paper describes a need for delegations to contain hashes in order to protect against malicous mirrors that may substitute one version of the delegated metadata with another. This cannot be detected without hashing of these files.

Closes #31

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -710,10 +710,10 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 * **4.4. File formats: snapshot.json**
 
-   The snapshot.json file is signed by the snapshot role. It lists the version
-   numbers, and optionally the size, of all metadata on the repository,
-   excluding root.json, timestamp.json and mirrors.json. For delegated roles,
-   the hash(es) are also listed.
+   The snapshot.json file is signed by the snapshot role.  It lists the version
+   numbers of all metadata on the repository, excluding root.json, timestamp.json and
+   mirrors.json.  The metadata length and hashes are OPTIONAL for the top-level and
+   all delegated targets roles.
 
    The "signed" portion of snapshot.json is as follows:
 
@@ -728,9 +728,8 @@ repo](https://github.com/theupdateframework/specification/issues).
 
        { METAPATH : {
              "version" : VERSION,
-             "length" : LENGTH,
-             "hashes" : HASHES,
-             ("custom" : { ... }) }
+             ("length" : LENGTH, |
+              "hashes" : HASHES) }
          , ...
        }
 
@@ -739,12 +738,12 @@ repo](https://github.com/theupdateframework/specification/issues).
 
    VERSION is listed for all roles available on the repository.
 
-   LENGTH is the optional integer length in bytes of the file. It is optional
-   for all roles.
+   LENGTH is the integer length in bytes of the metadata file. It is
+   OPTIONAL for all roles.
 
-   HASHES is a dictionary that specifies one or more hashes, including
-   the cryptographic hash function.  For example: { "sha256": HASH, ... }. It
-   is required for delegated roles, and optional for all others.
+   HASHES is the dictionary that specifies one or more hashes, including
+   the cryptographic hash function.  For example: { "sha256": HASH, ... }. It is
+   OPTIONAL for all roles.
 
    A snapshot.json example file:
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -710,10 +710,10 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 * **4.4. File formats: snapshot.json**
 
-   The snapshot.json file is signed by the snapshot role.  It lists the version
-   numbers of all metadata on the repository, excluding timestamp.json and
-   mirrors.json.  For the root role, the hash(es), size, and version number
-   are listed.
+   The snapshot.json file is signed by the snapshot role. It lists the version
+   numbers, and optionally the size, of all metadata on the repository,
+   excluding root.json, timestamp.json and mirrors.json. For delegated roles,
+   the hash(es) are also listed.
 
    The "signed" portion of snapshot.json is as follows:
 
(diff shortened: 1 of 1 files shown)


erickt:
Good afternoon, has anyone had a chance to review this patch? Is this the right way to implement section 5.6 of the [Mercury](https://www.usenix.org/system/files/conference/atc17/atc17-kuppusamy.pdf) paper?

tuf-spec.md
```diff
@@ -687,10 +687,10 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 * **4.4. File formats: snapshot.json**
 
-   The snapshot.json file is signed by the snapshot role.  It lists the version
-   numbers of all metadata on the repository, excluding timestamp.json and
-   mirrors.json.  For the root role, the hash(es), size, and version number
-   are listed.
+   The snapshot.json file is signed by the snapshot role. It lists the version
+   numbers, and optionally the size, of all metadata on the repository,
+   excluding root.json, timestamp.json and mirrors.json. For delegated roles,
```
trishankatdatadog: Hashes SHOULD be OPTIONAL for the top-level and all delegated targets roles.

tuf-spec.md
```diff
@@ -704,43 +704,58 @@ repo](https://github.com/theupdateframework/specification/issues).
    METAFILES is an object whose format is the following:
 
        { METAPATH : {
-             "version" : VERSION }
+             "version" : VERSION,
+             "length" : LENGTH,
+             "hashes" : HASHES,
```
trishankatdatadog: SHOULD be OPTIONAL, per comment above.

tuf-spec.md
```diff
@@ -704,43 +704,58 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+             ("custom" : { ... }) }
```
trishankatdatadog: There is no custom targets metadata for the snapshot role.

erickt:
Thanks for the review @trishankatdatadog! I incorporated your feedback.

tuf-spec.md
trishankatdatadog: No, hashes are optional now for the top-level targets and delegated targets roles. You might like to list hashes for a particular use case (see Section 5.6).

tuf-spec.md
```diff
@@ -688,9 +688,9 @@ repo](https://github.com/theupdateframework/specification/issues).
 * **4.4. File formats: snapshot.json**
 
    The snapshot.json file is signed by the snapshot role.  It lists the version
-   numbers of all metadata on the repository, excluding timestamp.json and
-   mirrors.json.  For the root role, the hash(es), size, and version number
-   are listed.
+   numbers of all metadata on the repository, excluding root.json, timestamp.json and
```
```suggestion
   numbers of only the top-level and all delegated targets role metadata.
```

tuf-spec.md
```diff
@@ -688,9 +688,9 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+   mirrors.json.  The metadata length and hashes are OPTIONAL for the top-level and
```
```suggestion
   The metadata length and hashes are OPTIONAL for the top-level and
```

tuf-spec.md
```diff
@@ -704,43 +704,57 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+              "hashes" : HASHES) }
          , ...
        }
 
    METAPATH is the metadata file's path on the repository relative to the
    metadata base URL.
 
-   VERSION is listed for the root file
-   and all other roles available on the repository.
+   VERSION is listed for all roles available on the repository.
```
```suggestion
   VERSION is listed for the top-level and all delegated targets roles available on the repository.
```

trishankatdatadog:
Can someone help to apply, commit, and merge the above 3 suggestions? Thanks! @lukpueh @JustinCappos

tuf-spec.md
lukpueh: Can we make this even clearer? I read your change as "top-level roles and delegated targets roles", which would mean all roles, which is not what we want. What about...

```suggestion
   numbers of only the top-level targets and all delegated targets role metadata.
```

tuf-spec.md
lukpueh: Same as above. 

```suggestion
   The metadata length and hashes are OPTIONAL for the top-level targets and
```

tuf-spec.md
lukpueh: Same as above. 

```suggestion
   VERSION is listed for the top-level targets and all delegated targets roles available on the repository.
```

lukpueh:
@trishankatdatadog, I'm happy to apply suggestions and merge. Let me know if my updates make sense to you.

trishankatdatadog:
@lukpueh Yes, please do apply, some GitHub bug prevents me from applying them, thanks

lukpueh:
@trishankatdatadog, I don't think it's a bug. We just don't have the permissions to commit to a branch on @erickt's fork. @erickt, would you mind applying the patches, so that we can merge?

trishankatdatadog:
@lukpueh Isn't there an on-by-default option that lets upstream edit PR?

lukpueh:
@trishankatdatadog Interesting, didn't know that. And an [API query to this PR page](https://api.github.com/repos/theupdateframework/specification/pulls/40) indeed returns `"maintainer_can_modify": true`. I'll just apply and merge locally.

lukpueh:
Interesting (II): I cannot apply commits but I can force push to @erickt's branch, which is what I just did in order to fix a conflict with #43, re-applying a there made keyid update, i.e. (`s/fce9cf.../66676d.../`) in the sample `snapshot.json` modified here.

I also added a commit to integrate @trishankatdatadog's and my suggestions (made via GitHub UI).

@trishankatdatadog, would you mind taking another look?

erickt:
Thanks all!

PR #45, opened 2019-06-18
Limit on the # of root metadata files downloaded

trishankatdatadog:
As per our discussion in theupdateframework/tuf#885

Cc @lukpueh

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1018,8 +1018,10 @@ repo](https://github.com/theupdateframework/specification/issues).
   by the authors of the application using TUF. For example, X may be tens of
   kilobytes. The filename used to download the root metadata file is of the
   fixed form VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1018,7 +1018,8 @@ repo](https://github.com/theupdateframework/specification/issues).
   by the authors of the application using TUF. For example, X may be tens of
   kilobytes. The filename used to download the root metadata file is of the
   fixed form VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is
-  not available, then go to step 1.8.
+  not available, or we have downloaded more than Y number of root metadata files,
+  then go to step 1.8.
```
lukpueh: Is it worth mentioning where Y comes from, similar to what we do in paragraph [*2. Download the timestamp metadata file*](https://github.com/theupdateframework/specification/pull/45/files#diff-94b2d37a08c587082c1ea1aaead6f973R1066)?

trishankatdatadog: Done, thanks @lukpueh

PR #46, opened 2019-07-01
Clarify consistent snapshots w/ regard to root rotation

trishankatdatadog:
Clarify how consistent snapshots are turned on and off during root metadata rotation.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1010,7 +1010,7 @@ repo](https://github.com/theupdateframework/specification/issues).
   somehow be able to establish a trusted line of continuity to the latest set
   of keys (see Section 6.1). To do so, the client MUST download intermediate
   root metadata files, until the latest available one is reached. Therefore, it
-  must temporarily turn on consistent snapshots.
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1009,7 +1009,8 @@ repo](https://github.com/theupdateframework/specification/issues).
   Since it may now be signed using entirely different keys, the client must
   somehow be able to establish a trusted line of continuity to the latest set
   of keys (see Section 6.1). To do so, the client MUST download intermediate
-  root metadata files, until the latest available one is reached.
+  root metadata files, until the latest available one is reached. Therefore, it
+  MUST temporarily turn on consistent snapshots.
```
lukpueh: Is the implication *"MUST turn on consistent snapshot to download intermediate files"* comprehensible to the target audience?

It might be helpful to mention the root filename version prefix, prescribed by consistent snapshots, and needed to increment over root files. But maybe it's not necessary.

trishankatdatadog: clarified, please lemme know if it helps

PR #47, opened 2019-08-01
Use hex encoding is used for ed25519 and target hashes

### LINKED ISSUE #42: divergence between implementations serializing ed25519 public keys
There is some ambiguity in the spec regarding how to encode ed25519's `public` portion of the `keyval`:

```
The 'ed25519' format is:

  { "keytype" : "ed25519",
    "scheme" : "ed25519",
    "keyval" : {"public" : PUBLIC}
  }

where PUBLIC is a 32-byte string.
```

While the ed25519 public key is a 32-byte string, I could find no implementation that's serializing the public key to JSON as an array of integers. Instead, implementations are serializing it into a JSON string, but there is a fork in how this is implemented:

*  [python-tuf](https://github.com/secure-systems-lab/securesystemslib/blob/beb3109d5bb462e5a60eed88fb40ed1167bd354e/securesystemslib/keys.py#L380) and [go-tuf](https://github.com/flynn/go-tuf/blob/master/data/types.go#L48) hex encode the public key
* [rust-tuf](https://github.com/heartsucker/rust-tuf/blob/develop/src/shims.rs#L266) encodes with base64url to encode the public key.
* [notary](https://github.com/docker/go/blob/master/canonical/json/encode.go#L53), and [hackage](https://github.com/haskell/hackage-security/blob/21a385c1a3c6222d4a7b6b8ecaff23431999d130/hackage-security/src/Hackage/Security/Key.hs#L215) base64 standard encode the public key.

Can you provide any guidance on what serialization format we should be using?

edit: turns out rust-tuf uses base64url instead of base64standard for encoding.

erickt:
According to #42, the spec should explicitly state that the ed25519 public key, and the target hash values are hex encoded values, since that is used by the majority of tuf implementations.

Closes #42

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -475,7 +475,8 @@ repo](https://github.com/theupdateframework/specification/issues).
 
           KEYID is the identifier of the key signing the ROLE dictionary.
 
-          SIGNATURE is a signature of the canonical JSON form of ROLE.
+          SIGNATURE is a hex-encoded signature of the canonical JSON form of
+          ROLE.
 
 
    All keys have the format:
@@ -532,7 +533,9 @@ repo](https://github.com/theupdateframework/specification/issues).
           "keyval" : {"public" : PUBLIC}
         }
 
-   where PUBLIC is a 32-byte string.
+   where:
+
+          PUBLIC is a 64-byte hex encoded string.
 
    The 'ecdsa' format is:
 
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
LGTM! Do you think we should clarify that signatures are in hex, too?

erickt:
@trishankatdatadog: it never hurts to remove ambiguity in a spec :) I updated the PR to also state the signature is hex-encoded.

PR #49, opened 2019-08-15
timestamp must only refer to the snapshot metadata

erickt:
@JustinCappos in https://github.com/heartsucker/rust-tuf/pull/207#issuecomment-521453860 recommended that the timestamp role should only allow a reference to the snapshot metadata, and no other roles.


JustinCappos:
@justincormack, I just wanted to double check there isn't any objection here.  I don't see a problem with this clarification.

justincormack:
SGTM

Issue #33, opened 2019-02-19
Status of 1.0 release of spec?


nealmcb:
I see that the [specification/README.rst](https://github.com/theupdateframework/specification/blob/master/README.rst) now no longer calls the spec a draft, but the [specification/tuf-spec.md](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md) itself (dated June 2018) still calls itself a draft.
* What is the status of the spec?
* Will there be a git tag when it is final?
* How does the version number relate to the tuf implementation versions and [tuf/ROADMAP.md](https://github.com/theupdateframework/tuf/blob/develop/docs/ROADMAP.md)?

awwad:
#### Spec Status:
The spec will continue to evolve, generally through [TAPs](https://github.com/theupdateframework/taps) (TUF Augmentation Proposals), meaning that future versions are expected.

We'll issue a git tag when each version of the spec is finalized.  I'll get back to you about the status of 1.0 in particular in the coming weeks.

As it predates this policy, version 0.9 of the spec -- which is finalized -- is available in all recent versions of the repository.


#### Spec Version in the Reference Implementation
The ways that specification versions are related to TUF implementation versions:
- Metadata [includes](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#4-document-formats) a specification version indicating the version of the spec to which it conforms.
- [Releases of the reference implementation](https://github.com/theupdateframework/tuf/releases) will make note of changes to the specification version with which they are intended to conform.

Do you think it would be helpful to additionally note conformance to a particular specification version in a comment in the code alongside TUF version, where it appears (e.g. in [setup.py](https://github.com/theupdateframework/tuf/blob/e97b0b091d9851baa3b7c5d31d758f0af20069d3/setup.py#L84))?

JustinCappos:
As for the spec itself, we're planning to wrap appropriate accepted TAPs
into the spec and then bump the version number.  We're figuring out the
cleanest way to do this now.

Justin

nealmcb:
Ahh - for convenience, here's a link to version 0.9 of the spec, from 2015-05-14: [specification/tuf-spec.0.9.txt](https://github.com/theupdateframework/specification/blob/master/tuf-spec.0.9.txt). I hadn't noticed that there were two versions of the spec next to each other in the repo.
The "1.0 (Draft)" version text has been used to describe the new draft since it first appeared in github on 11 October 2017, it seems.

[...]

PR #51, opened 2019-09-13
Introduce semantic versioning 2.0.0 and disambiguates spec versions

### LINKED ISSUE #36: Clarifying the format of "spec_version"
The spec defines the `spec_version"` as:

> SPEC_VERSION is the version number of the specification. Metadata is written according to version "spec_version" of the specification, and clients MUST verify that "spec_version" matches the expected version number. Adopters are free to determine what is considered a match (e.g., the version number exactly, or perhaps only the major version number (major.minor.fix).

Two requests:
* The document implies that `spec_version` is a string. Could this be explicitly mentioned?
* The document does not describe the format of the `spec_version` string. Can it contain arbitrary values, or is it only to be [semver](https://semver.org/) compatible?

Would it be acceptable to submit PRs to change the spec, or would these changes need to go through TAPs first?

lukpueh:
Closes #36 and  #33 

- Updates readme to mention use of semantic versioning for the specification
- Changes spec version "1.0" to "1.0.0", to follow semantic versioning, and bumps the "Last modified" date in the specification working document.
- Updates specification text to clarify that the `SPEC_VERSION` field in TUF metadata must be a string that follows semantic versioning.
- Moves `tuf-spec.0.9.txt` to `historical/tuf-spec.0.9.txt` to avoid confusion with current specification working document, i.e. `tuf-spec.md`, which currently is a draft for 1.0.0.


lukpueh:
> Do we need the "Draft" tag for version 1.0.0?

I'd only remove it if this revision is the final 1.0.0 version of the specification. In that case we should also tag it as such. Does anyone, more familiar with the specification, want to make the call? @JustinCappos, @mnm678, @SantiagoTorres, @trishankatdatadog?

README.rst
```diff
@@ -1,7 +1,7 @@
 The Update Framework specification
 ----------------------------------
 
-Latest: `version 1.0 <https://github.com/theupdateframework/specification/blob/master/tuf-spec.md>`_
+Latest: `version 1.0.0 (Draft) <https://github.com/theupdateframework/specification/blob/master/tuf-spec.md>`_
```
JustinCappos:
```suggestion
Latest: `version 1.0.0 <https://github.com/theupdateframework/specification/blob/master/tuf-spec.md>`_
```

tuf-spec.md
```diff
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **19 June 2018**
+Last modified: **13 September 2019**
 
-Version: **1.0 (Draft)**
+Version: **1.0.0 (Draft)**
```
```suggestion
Version: **1.0.0**
```

PR #52, opened 2019-10-03
Consistent snapshots for root metadata

trishankatdatadog:
discovery from theupdateframework/tuf#855


mnm678:
The requirement for consistent snapshots for root metadata is already included in section 5 of the specification (under client application step 1) where it states, "... it MUST temporarily turn on consistent snapshots in order to download versioned root metadata files as described next".

That being said, I think the addition here makes this requirement more clear.

PR #53, opened 2019-10-09
Remove links to outdated pep 458

lukpueh:
The first removed link is supposed to provide additional reasons for the need of consistent snapshots. Besides pointing to a removed anchor in an outdated and unmaintained version of pep 485, the spec already lists the same reasons that are mentioned in the pep.

The second link points to a different (existing) section in the same outdated document, but gives implementation details that IMO don't need to be in the spec. If wished we can update the second link instead of removing it.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1304,9 +1304,7 @@ non-volatile storage as FILENAME.EXT.
    high-level idea of the solution is that each snapshot will be contained in a
    so-called consistent snapshot. If a client is reading from one consistent
    snapshot, then the repository is free to write another consistent snapshot
-   without interrupting that client. For more reasons on why we need consistent
-   snapshots, please see
-   https://github.com/theupdateframework/pep-on-pypi-with-tuf#why-do-we-need-consistent-snapshots
+   without interrupting that client.
 
 * **7.1. Writing consistent snapshots**
 
@@ -1358,8 +1356,6 @@ non-volatile storage as FILENAME.EXT.
     released versions of root metadata files should always be provided
     so that outdated clients can update to the latest available root.
 
-    For more details on how this would apply on a community repository, please
-    see https://github.com/theupdateframework/pep-on-pypi-with-tuf#producing-consistent-snapshots
 
 * **7.2. Reading consistent snapshots**
 


tuf-spec.md
```diff
@@ -1358,8 +1356,6 @@ non-volatile storage as FILENAME.EXT.
     released versions of root metadata files should always be provided
     so that outdated clients can update to the latest available root.
 
-    For more details on how this would apply on a community repository, please
-    see https://github.com/theupdateframework/pep-on-pypi-with-tuf#producing-consistent-snapshots
```
lukpueh: As mentioned in the PR description, I'm happy to update this link (instead of removing it). I suggest to point to the official PEP, ie. https://www.python.org/dev/peps/pep-0458/#consistent-snapshots, which should soon have integrated our recent changes made in
https://github.com/secure-systems-lab/peps/blob/master/pep-0458.txt.

lukpueh:
I have a few more questions regarding consistent snapshots.

In [#L1320-L1323](https://github.com/theupdateframework/specification/blame/cf4561804154ea8ec9d36a0016b0fc5b7c86cfa8/tuf-spec.md#L1320-L1323):
> if the referrer metadata lists N cryptographic hashes of the referred file, then there must be N identical copies of the referred file, where each file will be distinguished only by the value of the digest in its filename.

This seems unpractical for large repositories. Do we really want to mandate it? What's the benefit of listing N cryptographic hashes, especially in the face of having N-times as many target files?

In [#L1330-L1334](https://github.com/theupdateframework/specification/blame/cf4561804154ea8ec9d36a0016b0fc5b7c86cfa8/tuf-spec.md#L1330-L1334):
> It is optional for an implementation to write identical copies at version_number.timestamp.json for record-keeping purposes, because a cryptographic hash of the timestamp metadata is usually not known in  advance.

Why should one write identical copies of timestamp and how does it help for record-keeping purposes?  Also, the "cryptographic hash" part seems a residue from "hash.timestamp.json"-times and should be replaced with "version", right?

In [#L1340-L1347](https://github.com/theupdateframework/specification/blame/cf4561804154ea8ec9d36a0016b0fc5b7c86cfa8/tuf-spec.md#L1340-L1347):
> Most importantly, no metadata file format must be updated to refer to the names of metadata or target files with their version numbers included. In other words, if a metadata file A refers to another metadata file B as filename.ext, then the filename must remain as filename.ext and not version_number.filename.ext. This rule is in place so that metadata signed by roles with offline keys will not be forced to sign for the metadata file whenever it is updated. In the next subsection, we will see how clients will reproduce the name of the intended file.

First of all, this paragraph only mentions version prefixes, when it should also mention hash prefixes.

Second of all, I'm not sure if the rationale here rule holds true. Doesn't any metadata A that refers to metadata B always include the version of metadata B, and must thus be re-signed on an update of B?  
Take for instance `timestamp`, which lists `"snapshot.json": {"version": 1, ...}`. Now if `snapshot` bumps to 2, `timestamp` has to change to `"snapshot.json": {"version": 2}` and re-sign, even though it ignores the change in the filename's prefix `(2.snapshot.json`).

Same goes for `snapshot` which lists `"root.json"`, `"targets.json"` and delegated targets along with their version numbers.

Issue #19, opened 2018-03-23
Prevent promiscuous delegations from violating security


awwad:
Items 4.5.1-4.5.9 in [the Detailed Workflows section (5.1) of the specification](https://github.com/theupdateframework/specification/blob/master/tuf-spec.md#5-detailed-workflows) do not instruct the implementer in how to "process" the delegated role files, and there are tempting dangers here that must be averted.

Consider this scenario (again... I know...):
- Targets delegates `/a/\*` to A
- Targets delegates `/b/\*` to B
- A delegates `/a/\*` to C, expecting threshold 1, key kA
- B delegates `/b/\*` to C, expecting threshold 1, key kB
- Targets, A, and B, are signed by all expected keys.
- C is signed by kA, but not by kB

Now, the client:
- updates top-level metadata, retrieving Root, Timestamp, Snapshot, and Targets.
- tries to get target info for target `/a/foo.txt`.
  - gets A.json
  - validates A.json based on the delegation info in Targets.json: valid
  - gets C.json
  - validates C.json based on the delegation info in A.json: valid, C.json was correctly signed by kA ((this step is not quite clear in the spec in 5.1:4.5.1+, but certainly implied))
- tries to get target info for target `/b/bar.txt`
  - gets B.json
  - validates B.json based on the delegation info in Targets.json: valid
  - regardless of whether or not C.json is cached, or whether we know that C.json has previously been validated (on a different delegation pathway, seeking a different target), MUST perform validation again, making sure that C.json was correctly signed by kB (and not just kA, as we validated before).

At that last step, it is likely that the client has cached C.json, kept it in a way that signals that it has been validated, in much the way that the top-level roles should be kept if and only if they are validated. The prior validation expecting kA cannot be allowed to pre-empt the validation expecting kB. This should probably be clear in the spec, as it is probably too easy to fail to catch. It requires that the implementer carefully consider the implications of promiscuous delegations.

Issue #32, opened 2019-02-11
How to verify delegated targets files


awwad:
Note that the issue identified [in Uptane Standard issue 36](https://github.com/uptane/uptane-standard/issues/36) applies to the TUF specification as well.  There is a tiny smidge more implied here in the TUF spec, but it still needs to be clarified.

PR #57, opened 2019-10-30
Adopt TAP 3 multi-role delegations metadata format

lukpueh:
Update section "4.5. File formats: targets.json and delegated target roles" to reflect metadata format introduced by [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md).

Note that TAP 3 is not implemented in the reference implementation, which claims to be spec 1.0.0 conformant!

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -789,7 +789,10 @@ repo](https://github.com/theupdateframework/specification/issues).
          "version" : VERSION,
          "expires" : EXPIRES,
          "targets" : TARGETS,
-         ("delegations" : DELEGATIONS)
+         ("keys_for_delegations" : {
+             KEYID : KEY,
+             ... },
+         "delegations" : [ DELEGATION, ... ])
        }
 
    TARGETS is an object whose format is the following:
@@ -824,27 +827,27 @@ repo](https://github.com/theupdateframework/specification/issues).
    TARGETPATH.  The application may use this information to guide download
    decisions.
 
-   DELEGATIONS is an object whose format is the following:
+   "keys_for_delegations" lists the public keys to verify signatures of
+   delegated targets roles. Revocation and replacement of delegated targets
+   roles keys is done by changing the keys in this field in the delegating
+   role's metadata.
 
-       { "keys" : {
-             KEYID : KEY,
-             ... },
-         "roles" : [{
-             "name": ROLENAME,
-             "keyids" : [ KEYID, ... ] ,
-             "threshold" : THRESHOLD,
-             ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
-              "paths" : [ PATHPATTERN, ... ]),
-             "terminating": TERMINATING,
-         }, ... ]
-       }
+   "delegations" is a list of DELEGATION objects whose format is the following:
 
-   "keys" lists the public keys to verify signatures of delegated targets roles.
-   Revocation and replacement of delegated targets roles keys is done by
-   changing the keys in this field in the delegating role's metadata.
+        {
+          "name": DELEGATION_NAME,
+          ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+          "paths" : [ PATHPATTERN, ... ]),
+          "terminating": TERMINATING,
+          "min_roles_in_agreement" : NUM_ROLES,
+          "roleinfo": [{
+            "rolename": ROLENAME,
+            "keyids": [ KEYID ],
+            "threshold": THRESHOLD,
+          }, ... ]
+        }
 
-   ROLENAME is the name of the delegated role.  For example,
-   "projects".
+   DELEGATION_NAME is the name of the delegation.
 
    TERMINATING is a boolean indicating whether subsequent delegations should be
    considered.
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -822,23 +825,27 @@ repo](https://github.com/theupdateframework/specification/issues).
...
 
-   ROLENAME is the name of the delegated role.  For example,
-   "projects".
+        {
+          "name": DELEGATION_NAME,
+          ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+          "paths" : [ PATHPATTERN, ... ]),
+          "terminating": TERMINATING,
+          "min_roles_in_agreement" : NUM_ROLES,
+          "roleinfo": [{
```
lukpueh: Prior to TAP 3 delegated roles were listed in the `roles` field. Should we keep that name?

tuf-spec.md
```diff
@@ -822,23 +825,27 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+   "delegations" is a list of DELEGATION objects whose format is the following:
 
-   ROLENAME is the name of the delegated role.  For example,
-   "projects".
+        {
+          "name": DELEGATION_NAME,
+          ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+          "paths" : [ PATHPATTERN, ... ]),
+          "terminating": TERMINATING,
+          "min_roles_in_agreement" : NUM_ROLES,
```
lukpueh: This was probably named `min_roles_in_agreement` to distinguish it from the signature `threshold` field. But I find it  odd that one threshold field  has such a specific name and the other one is generic. 

Maybe the scope (delegation vs. role) is disambiguation enough and we can both call `threshold`?

tuf-spec.md
```diff
@@ -790,7 +790,10 @@ repo](https://github.com/theupdateframework/specification/issues).
          "version" : VERSION,
          "expires" : EXPIRES,
          "targets" : TARGETS,
-         ("delegations" : DELEGATIONS)
+         ("keys_for_delegations" : {
```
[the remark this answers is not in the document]
lukpueh: IMHO this field should be under `delegations`, but then `delegations` would have to remain a dictionary and we'd have to find another name for the list that contains all the delegations...

tuf-spec.md
```diff
@@ -822,23 +825,27 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-         }, ... ]
-       }
+   "delegations" is a list of DELEGATION objects whose format is the following:
 
-   ROLENAME is the name of the delegated role.  For example,
-   "projects".
+        {
+          "name": DELEGATION_NAME,
+          ("path_hash_prefixes" : [ HEX_DIGEST, ... ] |
+          "paths" : [ PATHPATTERN, ... ]),
```
lukpueh: TAP 3 only mentions `paths`, but I guess we still want to support `path_hash_prefixes`?

lukpueh:
Interestingly, the client workflow had already mentioned multi-role delegation (prior to this PR). b46dc5e adds some clarification about the role threshold (`min_roles_in_agreement`).

trishankatdatadog:
Do we consider backwards-compatibility? If not, why not?

tuf-spec.md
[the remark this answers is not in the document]
mnm678: I think we should keep the keys separate for clarity.

tuf-spec.md
mnm678: maybe delegation_threshold?

tuf-spec.md
mnm678: I like `roles`, it makes it clear that there can be multiple roles

tuf-spec.md
```diff
@@ -1236,9 +1259,9 @@ non-volatile storage as FILENAME.EXT.
     of appearance.
 
       * **4.5.2.1**. If the current delegation is a multi-role delegation,
-      recursively visit each role, and check that each has signed exactly the
-      same non-custom metadata (i.e., length and hashes) about the target (or
-      the lack of any such metadata).
+      recursively visit each role, and check that a defined threshold of
```
mnm678: where does this threshold come from?

mnm678:
> Do we consider backwards-compatibility? If not, why not?

If yes, we would need to change the spec to version 2.0.0. However, we have previously stated that TAP 3 is part of TUF 1.0.0. I think for now we should include this change in 1.0.0.

tuf-spec.md
lukpueh: `min_roles_in_agreement`

tuf-spec.md
lukpueh: The field is a property of the `delegation` object. Addressing it as `delegation.delegation_threshold` seems like unnecessary redundancy.

tuf-spec.md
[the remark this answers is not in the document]
lukpueh: Okay, keeping it separate works for me. And what about my suggestion to call it `keys` only?  Because the `keys` field in `root.json` in reality also lists *"keys for delegations"*, i.e. keys to delegate trust to other top-level roles, but is only called `keys`.

lukpueh:
> Do we consider backwards-compatibility? If not, why not?

This PR is not the right place to ask these questions. This PR only restores consistency with backwards incompatible changes that have already been made. That is, claiming that the spec supports TAP 3 and TAP 4 (see #51). If 1.0.0 of the spec should not break backwards compatibility, then we must revert those claims.

But I agree with @mnm678 that we should get this (and TAP 4) into the spec now, and then make a formal 1.0.0 release. 1.0.0 will also help us to lead a more structured discussion about backwards compatibility (#33 has not yet been resolved!).

Independently, and I think that's the operational concern here (@trishankatdatadog?), we can discuss whether python-tuf should remain backwards compatible or not.

tuf-spec.md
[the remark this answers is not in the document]
mnm678: Calling it `keys` to match `root.json` makes sense.

tuf-spec.md
mnm678: I agree. Let's just call it `threshold`.

awwad: That was the reasoning, yes (distinguishing it from the other `threshold`).  If you think it'll not be confused, then it's nice to set it to something simple.

trishankatdatadog:
@mnm678 @lukpueh Understood. I wasn't aware that the 1.0.0 spec had made these claims about the TAPs, and that the client has not matched these claims yet. Good observation, and let's discuss in our meeting next week.

tuf-spec.md
lukpueh: Thinking about it some more, it does make sense to distinguish it from the other threshold. Ideally, we'd have equally specific names for both, e.g. `role.min_keys_in_agreement` and `delegation.min_roles_in_agreement` or `role.signature_threshold` and `delegation.role_threshold`.

 Is renaming the existing threshold field an option? Is it worthwhile? This will be a breaking change anyway. :)

Until then I'll just leave it as `threshold` and `min_roles_in_agreement`.

lukpueh:
I just pushed some renames (3c3b687) as discussed above. This discussion about `min_roles_in_agreement` is still open (see https://github.com/theupdateframework/specification/pull/57#discussion_r343733148). 

Also, should we propagate the renames to the actual TAP 3 document?

lukpueh:
Maybe now is also a good time to fix #19 and #32?

lukpueh:
As discussed in the TUF community meeting on  Dec 10, 2019 the changes of this PR require a major bump of the spec version (see [meeting notes about "Versioning" ](https://groups.google.com/forum/#!msg/theupdateframework/oPTliXibT_4/0tJ_POLGAwAJ)).

lukpueh:
Note: We will merge this PR into `draft`, because it is a major-type change of the spec. Bumping version number and last modified date will be done when PRing `draft` into `master`.

lukpueh:
> Could someone with write access merge this into draft (and possibly give me write access)?
Done. 

Thanks for the review!

PR #59, opened 2019-11-06
Fix link markup in README.rst

lukpueh:
Use restructured text instead of html syntax for links in acknowledgement section in README.rst. 

Here is how the document used to render [before](https://github.com/theupdateframework/specification/blob/0cddec0a60f95f06d2e23ebadbd876eeb62c1df3/README.rst) and how it renders [now (correctly)](https://github.com/lukpueh/specification/blob/511a15393289219456440182e952d5b39bf0af5e/README.rst).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -37,18 +37,16 @@ Acknowledgements
 ----------------
 
 This project is managed by the Linux Foundation under the Cloud Native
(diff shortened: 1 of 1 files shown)


SantiagoTorres:
LGTM. Thanks!

PR #61, opened 2019-11-14
Fix inconsistency in root key migration paragraph

### LINKED ISSUE #56: Ambiguity between section 5.1.3 and 6.1 regarding updating root keys
In section 5.1.3, the spec states:

> 1.3. Check signatures. Version N+1 of the root metadata file MUST have been signed by: (1) a threshold of keys specified in the trusted root metadata file (version N), and (2) a threshold of keys specified in the new root metadata file being validated (version N+1). If version N+1 is not signed as required, discard it, abort the update cycle, and report the signature failure. On the next update cycle, begin at step 0 and version N of the root metadata file.

However, this seems inconsistent with section 6.1 when speaking about recovering from a compromised root key:

> To replace a compromised root key or any other top-level role key, the root role signs a new root.json file that lists the updated trusted keys for the role. When replacing root keys, an application will sign the new root.json file with both the new and old root keys. Any time such a change is required, the root.json file is versioned and accessible by version number, e.g., 3.root.json. Clients update the set of trusted root keys by requesting the current root.json and all previous root.json versions, until one is found that has been signed by a threshold of keys that the client already trusts. This is to ensure that outdated clients remain able to update, without requiring all previous root keys to be kept to sign new root.json metadata.

As best as I can tell, 5.1 suggests that you update the root metadata from N, to N+1, N+2, to N+M etc. But 6.1 looks like it says you figure out the latest root version, and walk backwards from N+M, N+M-1, N+M-2, etc until you find a root metadata that's been signed by a quorum of keys trusted by the version N metadata.

In this case, it seems like we could have clients with different strategies could end up trusting different roots. For example:

* create a keys (K1) and sign root (R1) with them.
* roll the root metadata (R2) with new keys (K2) and sign the metadata with K1 and K2.
* we decide (R2) was compromised, so we generate (R3) and sign metadata with K1 and K3.

So a client thats updating from R1 with the forward updating updates to R2, and it will refuse to update to R3 because K2 didn't sign R3, but a client using the backwards strategy at R1 would skip R2 because R3 was signed by K1.

Looking at [python-tuf](https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L1105)  implements the forward strategy. Is that the correct approach? If so, could section 6.1 be updated to reflect this?

lukpueh:
Fixes #56 

Updates an outdated paragraph in section 6.1 that explained how root keys are rotated by backwards-verifying signatures from the latest root file available on the repository back to the latest root file available
on the client.

The correct (new) workflow, which is also described in step 1 of the client application workflow of section 5, performs verification in a forward-directed manner, i.e. starting at the latest root on the client and going forward to the latest root on the repo.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1292,22 +1292,23 @@ non-volatile storage as FILENAME.EXT.
    role.  When replacing root keys, an application will sign the new root.json
    file with both the new and old root keys. Any time such a change is
    required, the root.json file is versioned and accessible by version number,
-   e.g., 3.root.json. Clients update the set of trusted root keys by requesting
-   the current root.json and all previous root.json versions, until one is
-   found that has been signed by a threshold of keys that the client already
-   trusts. This is to ensure that outdated clients remain able to update,
-   without requiring all previous root keys to be kept to sign new root.json
-   metadata.
+   e.g., 3.root.json.
 
-   In the event that the keys being updated are root keys, it is important to
-   note that the new root.json must at least be signed by the keys listed as
-   root keys in the previous version of root.json, up to the threshold listed
-   for root in the previous version of root.json. If this is not the case,
-   clients will (correctly) not validate the new root.json file.  For example,
-   if there is a 1.root.json that has threshold 2 and a 2.root.json that has
-   threshold 3, 2.root.json MUST be signed by at least 2 keys defined in
-   1.root.json and at least 3 keys defined in 2.root.json. See step 1 in
-   Section 5.1 for more details.
+   Clients that have outdated root keys can update to the latest set of trusted
+   root keys, by incrementally downloading all intermediate root metadata
+   files, and verifying that each root metadata at a given version is signed by
+   a threshold of keys defined in the previous version of root, and a threshold
+   of keys defined in that given version of root.
+   For example, if there is a 1.root.json that has threshold 2 and a
+   2.root.json that has threshold 3, 2.root.json MUST be signed by at least 2
+   keys defined in 1.root.json and at least 3 keys defined in 2.root.json.
+   The client starts the root keys update update process with the latest
+   version of root metadata available on the client, and stops when no newer
+   root metadata is available from the repository. This ensures that an
+   outdated client can always correctly re-trace the chain of trust across
+   multiple root key updates, even if the latest set of root keys on the client
+   dates back multiple root metadata versions. See step 1 of the client
+   application workflow in Section 5 for more details.
 
    To replace a delegated developer key, the role that delegated to that key
    just replaces that key with another in the signed metadata where the
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1296,19 +1296,20 @@ non-volatile storage as FILENAME.EXT.
 
    Clients that have outdated root keys can update to the latest set of trusted
    root keys, by incrementally downloading all intermediate root metadata
-   files, and verifying that each root metadata at a given version is signed by
-   a threshold of keys defined in the previous version of root, and a threshold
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1292,22 +1292,23 @@ non-volatile storage as FILENAME.EXT.
...
+
+   Clients that have outdated root keys can update to the latest set of trusted
+   root keys, by incrementally downloading all intermediate root metadata
+   files, and verifying that each root metadata at a given version is signed by
+   a threshold of keys defined in the previous version of root, and a threshold
+   of keys defined in that given version of root.
+   For example, if there is a 1.root.json that has threshold 2 and a
+   2.root.json that has threshold 3, 2.root.json MUST be signed by at least 2
+   keys defined in 1.root.json and at least 3 keys defined in 2.root.json.
+   The client starts the root keys update update process with the latest
```
mnm678:
```suggestion
   The client starts the root key update process with the latest
```

tuf-spec.md
```diff
@@ -1292,22 +1292,23 @@ non-volatile storage as FILENAME.EXT.
...
-   clients will (correctly) not validate the new root.json file.  For example,
-   if there is a 1.root.json that has threshold 2 and a 2.root.json that has
-   threshold 3, 2.root.json MUST be signed by at least 2 keys defined in
-   1.root.json and at least 3 keys defined in 2.root.json. See step 1 in
-   Section 5.1 for more details.
+   e.g., 3.root.json.
+
+   Clients that have outdated root keys can update to the latest set of trusted
+   root keys, by incrementally downloading all intermediate root metadata
+   files, and verifying that each root metadata at a given version is signed by
```
```suggestion
   files, and verifying that each current version of the root metadata is signed by
```

tuf-spec.md
```diff
@@ -1292,22 +1292,23 @@ non-volatile storage as FILENAME.EXT.
...
+   a threshold of keys defined in the previous version of root, and a threshold
```
```suggestion
   a threshold of keys specified by its immediate predecessor as well as a threshold
```

tuf-spec.md
```diff
@@ -1292,22 +1292,23 @@ non-volatile storage as FILENAME.EXT.
...
+   of keys defined in that given version of root.
```
```suggestion
   of keys specified by itself.
```

lukpueh:
Thanks for your typo fix and rewordings, @mnm678 and @trishankatdatadog. I squashed them into one commit and added another one that mentions the possibility and limits of a freeze attack. Let me know what you think.

tuf-spec.md
```diff
@@ -1292,22 +1292,28 @@ non-volatile storage as FILENAME.EXT.
...
+   Clients that have outdated root keys can update to the latest set of trusted
+   root keys, by incrementally downloading all intermediate root metadata
+   files, and verifying that each current version of the root metadata is
+   signed by a threshold of keys specified by its immediate predecessor as well
+   as a threshold of keys specified by itself.
+   For example, if there is a 1.root.json that has threshold 2 and a
+   2.root.json that has threshold 3, 2.root.json MUST be signed by at least 2
+   keys defined in 1.root.json and at least 3 keys defined in 2.root.json. The
+   client starts the root key update process with the latest version of root
+   metadata available on the client, and stops when no newer root metadata is
```
```suggestion
   metadata available on the client, and stops when no version N+1 (where N is the latest trusted version) of the root metadata is
```

trishankatdatadog:
@lukpueh Sorry, I have one more suggestion. Other than that, it's great, thank you so much!

lukpueh:
It's a fine suggestion, @trishankatdatadog, thanks! Let me quickly re-fold the lines, squash the UI commit and force push.

PR #65, opened 2019-11-25
Remove problematic targets rollback attack check

lukpueh:
In the client application workflow, remove rollback attack check for top-level targets file, which is (1) redundant and (2) prevents recovery from a fast-forward attack.

(1) rollback attacks, via serving older versions of targets or top-level targets than the previously trusted versions, are already prevented by step 3.3.3 of the client workflow, where version numbers of targets and delegated targets in the new snapshot metadata are asserted to be greater than those in the prior trusted snapshot metadata. This, in combination with the 4.1 check that asserts that hashes and version of the actual targets metadata match the ones in the new trusted snapshot, makes another version number check, i.e the one removed in this PR, obsolete.

(2) fast-forward attack recovery, as described in 1.9, works by having the client remove the trusted timestamp and snapshot metadata after a non-root key rotation, so that the client can overcome the version comparison check, and update from a compromised high version to a recovered lower version. However, 1.9 does not mention removing trusted targets metadata after a key rotation. As a consequence, the additional version number check, removed in this PR, would prevent updating recovered targets metadata after a fast-forward attack.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1214,42 +1214,36 @@ non-volatile storage as FILENAME.EXT.
   trusted root metadata file.  If the new targets metadata file is not signed
   as required, discard it, abort the update cycle, and report the failure.
 
-  * **4.3**. **Check for a rollback attack.** The version number of the trusted
-  targets metadata file, if any, MUST be less than or equal to the version
-  number of the new targets metadata file.  If the new targets metadata file is
-  older than the trusted targets metadata file, discard it, abort the update
-  cycle, and report the potential rollback attack.
-
-  * **4.4**. **Check for a freeze attack.** The latest known time should be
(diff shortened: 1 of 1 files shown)


lukpueh:
An alternative fix for (2) would be updating the fast-forward recovery step to also untrust targets, as sketched out in https://github.com/theupdateframework/specification/commit/84103fcc726eece4fae3f7876a493bc1e988db3b.

I do, however, prefer the fix I proposed in this PR here, because it also addresses (1) and thereby makes the spec less complicated.

lukpueh:
Ping @JustinCappos. @mnm678 requested another review.

lukpueh:
As far as I understand @erickt is also in favor of merging this change (see https://github.com/theupdateframework/specification/issues/71).

PR #66, opened 2019-11-26
Add example for downloading foo/bar/baz target files

erickt:
This extends section 5.5.2 to include examples on how a client should download artifacts from a subdirectory. It uses the approach that python-tuf and go-tuf use, where downloading a target path like `path/to/file.ext` when consistent snapshots are enabled are fetched from `path/to/$HASH.file.ext`.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1265,10 +1265,12 @@ non-volatile storage as FILENAME.EXT.
   external program is used to compute the root hash of a tree of targets files,
   and this program does not provide the total size of all of these files.) If
   consistent snapshots are not used (see Section 7), then the filename used to
-  download the target file is of the fixed form FILENAME.EXT (e.g.,
-  foobar.tar.gz).  Otherwise, the filename is of the form HASH.FILENAME.EXT
+  download the target file is of the fixed form PATH/TO/FILENAME.EXT (e.g.,
+  foobar.tar.gz or baz/foo/bar).  Otherwise, the filename is of the form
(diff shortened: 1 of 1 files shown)


mnm678:
I agree with @lukpueh. This paragraph is referring just to the file, not the base url and path used to find the file. There might be a place to clarify how to download target files from a subdirectory elsewhere in the spec (maybe 3.1.1 or 4.5).

lukpueh:
@trishankatdatadog, do you disagree with the concerns I and @mnm678 raised? Do you think we should merge the PR as is?

lukpueh:
As per [tuf-spec.md#L408-L413](https://github.com/theupdateframework/specification/blame/3b744a9a799e8409038362e87f017b1826f4580e/tuf-spec.md#L408-L413)

>  **3.1.1 Target files**
   The filenames and the directory structure of target files available from
   a repository are not specified by the framework.  The names of these files
   and directories are completely at the discretion of the application using
   the framework.

I wonder if that paragraph deliberately ignores the fact that TUF does indeed specify filenames of target files, albeit only if "consistent snapshots" are used.

If we add @erickt's example to 3.1.1, we have to briefly mention consistent snapshots. I think it's worth it. What do others think?

trishankatdatadog:
We should resolve this PR, but someone needs to take over ownership

Cc @joshuagl @mnm678

joshuagl:
I'll take a stab at this next week.

PR #67, opened 2019-11-28
Update TARGETPATH and PATHPATTERN recommendations

### LINKED ISSUE #63: Is the TUF target path separator always "/"?
I noticed that in 4.4.5 spec defines as TARGETPATH:

> Each key of the TARGETS object is a TARGETPATH. A TARGETPATH is a path to a file that is relative to a mirror's base URL of targets. It should not have a leading path separator to avoid surprising behavior when constructing paths on disk.

Presumably this should be `/`, it says it's a file relative to the URL, and URLs use that as the path separator, but it'd be helpful if this was explicit.

This is relevant because in rust-tuf, we have a [PathTranslator](https://github.com/heartsucker/rust-tuf/blob/develop/src/client.rs#L87) in order to use alternative path separators to support converting `/` to `\` on windows, but I'm pretty sure this logic really should be moved to our FileSystemRepository.

lukpueh:
Closes #63 

In addition to recommending the omission of leading directory separators, this commit adds a recommendation for forward slashes as preferred directory separators in TARGETPATH and PATHPATTERN.

The rationale is that a TARGETPATH needs to be concatenated with a mirror's base URL, and as such it might yield unexpected behavior with non standardized paths. Same goes for Unix shell-style pattern matching of target paths using PATHPATTERN.

As suggested by @erickt this commit adds a reference to the "path-relative-URL string" definition from the WHATWG URL spec as the preferred TARGETPATH style.

This commit also switches to using the term "directory separator" instead of the prior used "path separator" to disambiguate from characters that separate paths (e.g. colon on Unix).

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -806,9 +806,13 @@ repo](https://github.com/theupdateframework/specification/issues).
        }
 
    Each key of the TARGETS object is a TARGETPATH.  A TARGETPATH is a path to
-   a file that is relative to a mirror's base URL of targets. It should not
-   have a leading path separator to avoid surprising behavior when constructing
-   paths on disk.
+   a file that is relative to a mirror's base URL of targets. To avoid
+   surprising behavior when resolving paths, it is RECOMMENDED that a
+   TARGETPATH uses the forward slash (/) as directory separator and does not
+   start with a directory separator. The recommendation for TARGETPATH aligns
+   with the ["path-relative-URL string"
+   definition](https://url.spec.whatwg.org/#path-relative-url-string) in the
+   WHATWG URL specification.
 
    It is allowed to have a TARGETS object with no TARGETPATH elements.  This
    can be used to indicate that no target files are available.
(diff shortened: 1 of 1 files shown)


lukpueh:
Two questions remain:

1. Do we want to add a note that says that tuf repository and client software should adapt accordingly if the local system uses other path styles.
--> IMO it's not necessary (goes without saying) 

2. Do we really need to reference the WHATWG URL spec, and if so, is ["path-relative-URL"](https://url.spec.whatwg.org/#path-relative-url-string)  the right choice, or should it rather be ["path-relative-scheme-less-URL"](https://url.spec.whatwg.org/#path-relative-scheme-less-url-string).
--> IMO it's fine as I proposed it in the PR. I think the reference is helpful but we don't need to be more specific than that, i.e. recommend against a scheme prefix (e.g. file://, etc...)

Thoughts, @erickt, @trishankatdatadog, @mnm678, @JustinCappos?

mnm678:
The fix looks good. As far as the remaining questions:

1: This seems clear from the existing text. We can include more detail about adapting the paths to a local system in a POUF.

2: The reference to the URL spec is helpful, but not necessary. TUF could theoretically work with non-standard urls as long as clients and servers agree. That said, I would leave it in as a helpful reference for implementers.

erickt:
This also looks good for me, thanks for writing this up. I'd love for this to also be a requirement for maximum interoperability, but I can understand that might be difficult since we might already have some TUF-1.0 users in the field that might be using alternative separators.

lukpueh:
Thanks for the feedback, @mnm678, @erickt and @trishankatdatadog!

Issue #71, opened 2019-12-04
Ambiguity around rotating keys and deleting metadata


erickt:
In section 5.1.9, it states:

> 1.9. If the timestamp and / or snapshot keys have been rotated, then delete the trusted timestamp and snapshot metadata files. This is done in order to recover from fast-forward attacks after the repository has been compromised and recovered. A fast-forward attack happens when attackers arbitrarily increase the version numbers of: (1) the timestamp metadata, (2) the snapshot metadata, and / or (3) the targets, or a delegated targets, metadata file in the snapshot metadata. Please see the Mercury paper for more details.

There some ambiguity here:

* We should define what it means to rotate a key. Do we only remove the metadata if we add a new key and remove an old key at the same time in a root metadata? Or should we also delete the local metadata if we remove a key?
* In the case where we have multiple keys and a threshold > 1 for a role, do we still delete the metadata if we only rotate one key? Theoretically, it should be safe to rotate less than the threshold number of keys in a given root metadata, since an attacker shouldn't be able to perform a fast-forward attack with less than the threshold number of compromised keys.
* Why do we delete both the local timestamp and snapshot metadata if just the timestamp key, rather than just deleting the timestamp metadata? Similarly for just a snapshot key rotation. Does this protect against a known attack, or is this more about cleaning up attacker controlled files that might contain other unknown attacks?
* Why do we not delete the targets metadata if those keys are rotated? (I think this might be addressed though by @lukpueh in #65 or https://github.com/theupdateframework/specification/commit/84103fcc726eece4fae3f7876a493bc1e988db3b if that gets merged in).

Thanks again!

mnm678:
I opened #74 to address the first two points you made and clarify what rotation means in this context, please review/comment. I will leave the last point to @lukpueh in #65. If that does not answer the question @erickt, we can move the discussion there.

As for your third point, the trusted snapshot metadata will always need to be deleted after a fast-forward attack because it will list the fast-forwarded version of the timestamp or other metadata. I am not sure that this holds for the timestamp (if only snapshot is compromised, the timestamp metadata can still be trusted). However, if online keys are used for both of these roles, in practice it would make sense to rotate (revoke and replace) both timestamp and snapshot in the event of a compromise. @trishankatdatadog @JustinCappos is there another attack that is prevented by deleting both timestamp and snapshot? Either way, I think it would be safer to continue to delete both the timestamp and snapshot metadata as part of compromise recovery.

lukpueh:
[...]

PR #72, opened 2019-12-07
Clarify how delegated roles are downloaded

erickt:
Section 5.4.5 is a little vague how on delegated targets are fetched and validated. This updates that section to use the same logic and verification process as downloading the top-level targets role to be explicit.

One thing to point out though is that the old section 4.5 suggests that we don't report verification errors to the user. I've preserved this in my explicit version. I imagine users would still be notified if their delegated roles may be undergoing an attack. Is this intentional, or should I switch to the "abort the update cycle, and report the potential rollback attack"-style phrasing used elsewhere in the spec?

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1235,9 +1235,8 @@ non-volatile storage as FILENAME.EXT.
   and report the potential freeze attack.
 
   * **4.5**. **Perform a preorder depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.**  Note: If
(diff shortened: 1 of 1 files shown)


mnm678:
I agree that we should report this error like other places in the specification.

lukpueh:
@erickt, I took the liberty to take over this PR, addressing my own review comments (typos, missing freeze attack check) and your question regarding error reporting.

I submitted a new PR under #86. It also adds work from me and @mnm678 to address the concerns you raised in #71.

It would be great if you could take a look. Regardless, I'd like to thank you for your great contributions.

PR #74, opened 2019-12-13
Clarify rotation in 1.9

mnm678:
Addresses part of #71 

Replace the use of 'rotate' with a description of the keys that would be removed when recovering from a fast-forward attack.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1116,7 +1116,8 @@ repo](https://github.com/theupdateframework/specification/issues).
   cycle, report the potential freeze attack.  On the next update cycle, begin
   at step 0 and version N of the root metadata file.
 
-  * **1.9**. **If a threshold of timestamp and / or snapshot keys have been removed, then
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1116,7 +1116,7 @@ repo](https://github.com/theupdateframework/specification/issues).
   cycle, report the potential freeze attack.  On the next update cycle, begin
   at step 0 and version N of the root metadata file.
 
-  * **1.9**. **If the timestamp and / or snapshot keys have been rotated, then
+  * **1.9**. **If a threshold of timestamp and / or snapshot keys have been removed, then
```
trishankatdatadog: A threshold of timestamp / snapshot keys compared to what? Version N of the root metadata file from Step 1.1?

mnm678: Yes, I will clarify in the text.

lukpueh:
Superseded by #86

PR #84, opened 2020-01-08
Add description of ptt

mnm678:
Addresses [tuf#127](https://github.com/theupdateframework/tuf/issues/127)

Adds description of permanent takedown threat to the malicious mirror attack. This attack is one way in which an attacker could compromise a mirror.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -216,7 +216,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
       * This version (1.0.0) of the specification adheres to the following TAPS:
 
-        - [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md):
+        - [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md): 
             Multi Role Delegations
         - [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md):
             Multiple Repository Consensus on entrusted targets
@@ -1389,7 +1389,7 @@ non-volatile storage as FILENAME.EXT.
     snapshots are not written by the repository, then the attribute may either
     be left unspecified or be set to the False value.  Otherwise, it must be
     set to the True value.
-
+    
     Regardless of whether consistent snapshots are ever used or not, all
     released versions of root metadata files should always be provided
     so that outdated clients can update to the latest available root.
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -173,7 +173,9 @@ repo](https://github.com/theupdateframework/specification/issues).
       of the problem.
 
       + **Malicious mirrors preventing updates.**  A repository mirror cannot
-      prevent updates from good mirrors.
+      prevent updates from good mirrors. This includes mirrors that have been
```
trishankatdatadog: Could you elaborate on how bad mirrors could prevents updates from good mirrors (i.e., what are the plausible mechanisms)?

Also, do you mean good mirrors by those "that have been taken down due to" etc?

tuf-spec.md
```diff
@@ -173,9 +173,10 @@ repo](https://github.com/theupdateframework/specification/issues).
       of the problem.
 
       + **Malicious mirrors preventing updates.**  A repository mirror cannot
-      prevent updates from good mirrors. This includes mirrors that have been
-      taken down due to a DoS attack, malware, or pressure on administrators
-      (known as a permanent takedown threat).
+      prevent updates from good mirrors (such as by sharing malicious files).
```
trishankatdatadog: Sorry, I'm still confused. How can bad mirror prevent updates from good mirrors by sharing malicious files?

tuf-spec.md
```diff
@@ -173,9 +173,10 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+      Malicious mirrors include mirrors that have been taken down due to a
+      DoS attack, malware, or pressure on administrators (known as a permanent
+      takedown threat).
```
trishankatdatadog: Sorry, but I don't think you mean "taken down" for DoS attacks _and_ malware or pressure on administrators?

mnm678:
@trishankatdatadog My aim was to describe Permanent Takedown Threat as a subset of Malicious mirror attacks. However, I think this description may be adding more confusion than clarity. I can either:

* Make PTT it's own section in the attack descriptions. This would give space to describe this attack more clearly, but it might be redundant with the description of malicious mirrors and DoS attacks.
* Remove this description. The descriptions of malicious mirror attacks and DoS attacks have been added since the issue was opened, and could cover PTT without any additional text.
* Keep working on the text here to make it clear that PTT refers to attacks on servers and domains.

I am leaning toward the second option, but will work on a better way to describe PTT if others think it's an important addition to this section of the specification

trishankatdatadog:
Second option sounds good. Thanks @mnm678!

PR #85, opened 2020-01-22
Fast-forward recovery and rollback prevention

mnm678:
Addresses part of #71 

Clarify how and when to delete targets, snapshot, and timestamp metadata to recover from a fast-forward attack. In addition, adds an additional rollback check to prevent a rollback attack in the event of a timestamp compromise.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -214,7 +214,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
       * This version (1.0.0) of the specification adheres to the following TAPS:
 
-        - [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md): 
+        - [TAP 3](https://github.com/theupdateframework/taps/blob/master/tap3.md):
             Multi Role Delegations
         - [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md):
             Multiple Repository Consensus on entrusted targets
@@ -1119,7 +1119,7 @@ repo](https://github.com/theupdateframework/specification/issues).
   * **1.9**. **Fast-forward attack recovery** A _fast-forward attack_ happens
   when attackers arbitrarily increase the version numbers of: (1) the timestamp
   metadata, (2) the snapshot metadata, and / or (3) the targets, or a delegated
-  targets, metadata file in the snapshot metadata. To recover from fast-forward
+  targets, metadata file. To recover from fast-forward
   attacks after the repository has been compromised and recovered, certain
   metadata files need to be deleted as specified in this section. Please see
   [the Mercury paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1261,8 +1261,11 @@ non-volatile storage as FILENAME.EXT.
 
       * **4.5.2.1**. **Fast-forward attack recovery** If a threshold of
       delegated targets keys for the current delegation are removed from the
-      delegating targets metadata, delete the old delegated targets metadata for
-      the current delegation along with the snapshot and timestamp metadata.
+      delegating targets metadata (either the top-level targets or delegated
+      targets metadata that provided the current delegation), delete the old
+      delegated targets metadata for the current delegation along with the
+      snapshot and timestamp metadata. Skip this step if the current delegation
+      is the top-level targets role.
 
       * **4.5.2.1**. If the current delegation is a multi-role delegation,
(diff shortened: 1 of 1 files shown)


mnm678:
@lukpueh Thanks for the review! I addressed your comments.

tuf-spec.md
```diff
@@ -1116,16 +1116,26 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+  when attackers arbitrarily increase the version numbers of: (1) the timestamp
+  metadata, (2) the snapshot metadata, and / or (3) the targets, or a delegated
+  targets, metadata file. To recover from fast-forward
+  attacks after the repository has been compromised and recovered, certain
+  metadata files need to be deleted as specified in this section. Please see
+  [the Mercury paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
   for more details.
 
+    * **1.9.1**. **Targets recovery** If a threshold of targets keys are removed
+    from the root metadata, delete the old targets, snapshot, and timestamp
```
trishankatdatadog: I think it's fairly clear, but you could make it clearer that you mean the top-level targets file, and not any delegated targets file?

tuf-spec.md
```diff
@@ -1222,36 +1232,47 @@ non-volatile storage as FILENAME.EXT.
...
+  cycle, and report the potential rollback attack.
+
+  * **4.4**. **Check for a freeze attack.** The latest known time should be
   lower than the expiration timestamp in the new targets metadata file.  If so,
   the new targets metadata file becomes the trusted targets metadata file.  If
   the new targets metadata file is expired, discard it, abort the update cycle,
   and report the potential freeze attack.
 
-  * **4.4**. **Perform a preorder depth-first search for metadata about the
+  * **4.5**. **Fast-forward attack recovery** If a threshold of delegated
```
trishankatdatadog: Good check, but we can't do this before the subsequent preorder DFS, no?

mnm678: Good catch, this check should happen during the DFS, maybe as 4.6.2.1

lukpueh: Good catch indeed, @trishankatdatadog. Do you guys think it is clear that the "delegating targets metadata" (as now mentioned in 4.5.2.1) can also be the top-level targets metadata? Or should we add a clarification?

lukpueh:
@mnm678,  I took the liberty to untangle the commit history a bit. It seems like there are a bunch of duplicate commits from this PR and also from prior PRs due to a merge. 

So I undid your merge by hard resetting to one of the merge commits' parents (i.e. db9b54b24399c840405ca50eef509df5dab1a46b), and cherry-picked the commit that was missing (i.e. c2aa5152977b66bac991250566a39c0650ec6da5) onto it. I also did some interactive rebasing to clean up the commit messages and squash in the one UI commit.

You can replace your branch with my cleaned up version with these commands:

```bash
# Fetch my version of your branch
git fetch git@github.com:lukpueh/specification.git fast_forward-lp:fast_forward-lp

# Go to your branch (if you're not already on it)
git checkout fast_forward

# Optionally diff it with mine to see that it's indeed the same contents
git diff fast_forward-lp

# Replace yours with mine
git reset --hard fast_forward-lp

# Force push to update the branch on GitHub
# (you might have to change "origin" to whatever points to git@github.com:mnm678/specification.git)
git push origin fast_foward -f
```

tuf-spec.md
trishankatdatadog: Always better to clarify, in my humble experience!

tuf-spec.md
```diff
@@ -1222,36 +1232,47 @@ non-volatile storage as FILENAME.EXT.
...
 
-    * **4.4.2**. Otherwise, recursively search the list of delegations in order
+    * **4.5.2**. Otherwise, recursively search the list of delegations in order
     of appearance.
 
-      * **4.4.2.1**. If the current delegation is a multi-role delegation,
+      * **4.5.2.1**. **Fast-forward attack recovery** If a threshold of
+      delegated targets keys for the current delegation are removed from the
+      delegating targets metadata, delete the old delegated targets metadata for
+      the current delegation along with the snapshot and timestamp metadata.
```
trishankatdatadog: Why the snapshot and timestamp again here?

mnm678: If there was a fast forward attack, the snapshot would contain the fast-forwarded version number of the delegated targets file

tuf-spec.md
mnm678: I agree, it's better to be clear. I'll add a note

lukpueh:
Superseded by #86

PR #86, opened 2020-01-30
Clarify rollback attack prevention and fast-forward attack recovery

### LINKED ISSUE #71: Ambiguity around rotating keys and deleting metadata

lukpueh:
This PR combines and updates #72, which clarifies the verification routines for delegated targets, and #74 and #85, which fix #71, i.e. resolve ambiguity around rotating keys and deleting metadata, in order to recover from fast-forward attacks, by temporarily disabling rollback attack checks.

**Pre-requisite**
- 0ecf980 reverts #65 (see commit message for rationale)

**Takeover of #72**
- 75ac163 is a verbatim copy of the sole commit from #72 courtesy of @erickt
- 329361a,  c36e0c0  and e4a6f9e address #72-reviews (by [*@lukpueh*](https://github.com/theupdateframework/specification/pull/72#pullrequestreview-331245904) and [*@mnm678*](https://github.com/theupdateframework/specification/pull/72#issuecomment-565595339))

**Takeover of #85**
- Please see commit messages of b677f6a, fa2d263 and most notably 65e042d for details

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1220,14 +1220,13 @@ non-volatile storage as FILENAME.EXT.
     metadata file, discard it, abort the update cycle, and report the potential
     rollback attack.
 
-    * **3.3.2**. The version number of the targets metadata file, and all
-    delegated targets metadata files (if any), in the trusted snapshot metadata
-    file, if any, MUST be less than or equal to its version number in the new
-    snapshot metadata file. Furthermore, any targets metadata filename that was
-    listed in the trusted snapshot metadata file, if any, MUST continue to be
-    listed in the new snapshot metadata file.  If any of these conditions are
-    not met, discard the new snapshot metadadata file, abort the update cycle,
-    and report the failure.
+    * **3.3.2**. The version number of the top-level targets metadata file, in
+    the trusted snapshot metadata file, if any, MUST be less than or equal to
+    its version number in the new snapshot metadata file. Furthermore, any
+    targets metadata filename that was listed in the trusted snapshot metadata
+    file, if any, MUST continue to be listed in the new snapshot metadata file.
+    If any of these conditions are not met, discard the new snapshot metadadata
+    file, abort the update cycle, and report the failure.
 
   * **3.4**. **Check for a freeze attack.** The latest known time should be
   lower than the expiration timestamp in the new snapshot metadata file.  If
@@ -1287,7 +1286,18 @@ non-volatile storage as FILENAME.EXT.
       * **4.5.2.1**. Let DELEGATE denote the current target role TARGETS is
       delegating to.
 
-      * **4.5.2.2**. **Download the DELEGATE targets metadata file**, up to either
+      * **4.5.2.2**. **Fast-forward attack recovery.** If a threshold of
+      delegated targets keys for the current delegation are removed from the
+      TARGETS metadata, delete the trusted DELEGATE metadata, if any, and the
+      previously trusted snapshot metadata.
+
+      * **4.5.2.3**. **Check for a rollback attack via snapshot.** The version number of the
+      DELEGATE metadata in the previous trusted snapshot metadata, if any, MUST
+      be less than or equal to its version number in the new trusted snapshot
+      metadata. If this is not the case, abort the update cycle, and report the
+      potential rollback attack.
+
+      * **4.5.2.4**. **Download the DELEGATE targets metadata file**, up to either
       the number of bytes specified in the snapshot metadata file, or some Z
       number of bytes. The value for Z is set by the authors of the application
       using TUF. For example, Z may be tens of kilobytes. IF DELEGATE cannot be
@@ -1300,39 +1310,40 @@ non-volatile storage as FILENAME.EXT.
       in the snapshot metadata file.  In either case, the client MUST write the
       file to non-volatile storage as FILENAME.EXT.
 
-      * **4.5.2.3**. **Check against snapshot metadata.** The hashes (if any), and
+      * **4.5.2.5**. **Check against snapshot metadata.** The hashes (if any), and
       version number of the new DELEGATE metadata file MUST match the trusted
-      snapshot metadata.  This is done, in part, to prevent a mix-and-match
+      snapshot metadata, if any.  This is done, in part, to prevent a mix-and-match
       attack by man-in-the-middle attackers. If the new DELEGATE metadata file
       does not match, abort the update cycle, and report the failure.
 
-      * **4.5.2.4**. **Check for an arbitrary software attack.** The new DELEGATE
+      * **4.5.2.6**. **Check for an arbitrary software attack.** The new DELEGATE
       metadata file MUST have been signed by a threshold of keys specified in the
       TARGETS metadata file.  If the new DELEGATE metadata file is not signed
       as required, abort the update cycle, and report the failure.
 
-      * **4.5.2.5**. **Check for a rollback attack.** The version number of the
-      trusted DELEGATE metadata file, if any, MUST be less than or equal to the
-      version number of the new DELEGATE metadata file.  If the new DELEGATE
-      metadata file is older than the trusted DELEGATE metadata file, discard
-      it, abort the update cycle, and report the potential rollback attack.
+      * **4.5.2.7**. **Check for a rollback attack on the DELEGATE metadata.**
+      The version number of the trusted DELEGATE metadata file, if any, MUST be
+      less than or equal to the version number of the new DELEGATE metadata
+      file.  If the new DELEGATE metadata file is older than the trusted
+      DELEGATE metadata file abort the update cycle, and report the potential
+      rollback attack.
 
-      * **4.5.2.6**. **Check for a freeze attack.** The latest known time
+      * **4.5.2.8**. **Check for a freeze attack.** The latest known time
       should be lower than the expiration timestamp in the new DELEGATE
       metadata file. If so, the new DELEGATE file becomes the trusted DELEGATE
       file. If the new DELEGATE metadata file is expired, abort the update
       cycle, and report the potential freeze attack.
 
-      * **4.5.2.7**. If the current delegation is a multi-role delegation,
+      * **4.5.2.9**. If the current delegation is a multi-role delegation,
       recursively visit each role, and check that each has signed exactly the
       same non-custom metadata (i.e., length and hashes) about the target (or
       the lack of any such metadata). Otherwise, abort the update cycle, and
       report the failure.
 
-      * **4.5.2.8**. If the current delegation is a terminating delegation,
+      * **4.5.2.10**. If the current delegation is a terminating delegation,
       then jump to step 5.
 
-      * **4.5.2.9**. Otherwise, if the current delegation is a non-terminating
+      * **4.5.2.11**. Otherwise, if the current delegation is a non-terminating
       delegation, continue processing the next delegation, if any, by repeating
       step 4.5 with DELEGATE as the current TARGET role. Stop the search, and
       jump to step 5 as soon as a delegation returns a result.
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1274,14 +1274,14 @@ non-volatile storage as FILENAME.EXT.
       * **4.5.2.1**. Let DELEGATE denote the current target role TARGETS is
       delegating to.
 
-      * **4.5.2.2**. **Download the DELEGATE tarets metadata file**, up to either
+      * **4.5.2.2**. **Download the DELEGATE targets metadata file**, up to either
       the number of bytes specified in the snapshot metadata file, or some Z
       number of bytes. The value for Z is set by the authors of the application
       using TUF. For example, Z may be tens of kilobytes. IF DELEGATE cannot be
       found, end the search and report the target cannot be found.  If
       consistent snapshots are not used (see Section 7), then the filename used
       to download the targets metadata file is of the fixed form FILENAME.EXT
-      (e.g., delegated_rol.json).  Otherwise, the filename is of the form
+      (e.g., delegated_role.json).  Otherwise, the filename is of the form
       VERSION_NUMBER.FILENAME.EXT (e.g., 42.delegated_role.json), where
       VERSION_NUMBER is the version number of the DELEGATE metadata file listed
       in the snapshot metadata file.  In either case, the client MUST write the
@@ -1303,7 +1303,7 @@ non-volatile storage as FILENAME.EXT.
       * **4.5.2.5**. **Check for a rollback attack.** The version number of the
       trusted DELEGATE metadata file, if any, MUST be less than or equal to the
       version number of the new DELEGATE metadata file.  If the new DELEGATE
-      `metadata file is older than the trusted DELEGATE metadata file, discard
+      metadata file is older than the trusted DELEGATE metadata file, discard
       it, end the search, and report the target cannot be found.
 
       * **4.5.2.6**. If the current delegation is a multi-role delegation,
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1135,12 +1135,15 @@ repo](https://github.com/theupdateframework/specification/issues).
 
   * **1.9**. **Fast-forward attack recovery** A _fast-forward attack_ happens
   when attackers arbitrarily increase the version numbers in any of the
-  timestamp, snapshot, targets, or delegated targets metadata. To recover from
+  timestamp, snapshot, targets, or delegated targets metadata. The attacker goal
+  is to cause clients to refuse to update the metadata later because the attacker's
+  listed metadata version number (possibly MAX_INT) is greater than the new valid 
+  version.  To recover from
   fast-forward attacks after the repository has been compromised and recovered,
   certain metadata files need to be deleted as specified in this section.
   Please see [the Mercury
   paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
-  for more details.
+  for more details on fast-forward attacks.
 
     * **1.9.1**. **Targets recovery** If a threshold of targets keys have been
     removed in the new trusted root metadata compared to the previous trusted
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1142,9 +1142,10 @@ repo](https://github.com/theupdateframework/specification/issues).
   paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
   for more details.
 
-    * **1.9.1**. **Targets recovery** If a threshold of targets keys are removed
-    from the root metadata, delete the old top-level targets, snapshot, and
-    timestamp metadata files.
+    * **1.9.1**. **Targets recovery** If a threshold of targets keys have been
+    removed in the new trusted root metadata compared to the previous trusted
+    root metadata, delete the old top-level targets and snapshot metadata
+    files.
 
     * **1.9.2**. **Snapshot recovery** If a threshold of snapshot keys have
     been removed in the new trusted root metadata compared to the previous
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1306,16 +1306,22 @@ non-volatile storage as FILENAME.EXT.
       metadata file is older than the trusted DELEGATE metadata file, discard
       it, end the search, and report the target cannot be found.
 
-      * **4.5.2.6**. If the current delegation is a multi-role delegation,
+      * **4.5.2.6**. **Check for a freeze attack.** The latest known time
+      should be lower than the expiration timestamp in the new DELEGATE
+      metadata file. If so, the new DELEGATE file becomes the trusted DELEGATE
+      file. If the new DELEGATE metadata file is expired, discard it, end the
+      search, and report the target cannot be found.
+
+      * **4.5.2.7**. If the current delegation is a multi-role delegation,
       recursively visit each role, and check that each has signed exactly the
       same non-custom metadata (i.e., length and hashes) about the target (or
       the lack of any such metadata). Otherwise, discard it, end the search,
       and report the target cannot be found.
 
-      * **4.5.2.7**. If the current delegation is a terminating delegation,
+      * **4.5.2.8**. If the current delegation is a terminating delegation,
       then jump to step 5.
 
-      * **4.5.2.8**. Otherwise, if the current delegation is a non-terminating
+      * **4.5.2.9**. Otherwise, if the current delegation is a non-terminating
       delegation, continue processing the next delegation, if any, by repeating
       step 4.5 with DELEGATE as the current TARGET role. Stop the search, and
       jump to step 5 as soon as a delegation returns a result.
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1258,9 +1258,8 @@ non-volatile storage as FILENAME.EXT.
   and report the potential freeze attack.
 
   * **4.5**. **Perform a preorder depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.**  Note: If
-  any metadata requested in steps 4.5.1 - 4.5.2.3 cannot be downloaded nor
-  validated, end the search and report that the target cannot be found.
+  desired target.** Let TARGETS be the current metadata, beginning with the
+  top-level targets metadata role.
 
     * **4.5.1**. If this role has been visited before, then skip this role (so
     that cycles in the delegation graph are avoided).  Otherwise, if an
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1116,16 +1116,28 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-  the snapshot metadata, and / or (3) the targets, or a delegated targets,
-  metadata file in the snapshot metadata. Please see [the Mercury
+  * **1.9**. **Fast-forward attack recovery** A _fast-forward attack_ happens
+  when attackers arbitrarily increase the version numbers in any of the
+  timestamp, snapshot, targets, or delegated targets metadata. To recover from
+  fast-forward attacks after the repository has been compromised and recovered,
+  certain metadata files need to be deleted as specified in this section.
+  Please see [the Mercury
   paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
   for more details.
```
joshuagl:
```suggestion
  for more details on fast-forward attacks.
```

tuf-spec.md
```diff
@@ -1116,16 +1116,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-  * **1.9**. **If the timestamp and / or snapshot keys have been rotated, then
-  delete the trusted timestamp and snapshot metadata files.** This is done in
-  order to recover from fast-forward attacks after the repository has been
-  compromised and recovered. A _fast-forward attack_ happens when attackers
-  arbitrarily increase the version numbers of: (1) the timestamp metadata, (2)
-  the snapshot metadata, and / or (3) the targets, or a delegated targets,
-  metadata file in the snapshot metadata. Please see [the Mercury
+  * **1.9**. **Fast-forward attack recovery** A _fast-forward attack_ happens
+  when attackers arbitrarily increase the version numbers in any of the
+  timestamp, snapshot, targets, or delegated targets metadata. To recover from
```
```suggestion
  timestamp, snapshot, targets, or delegated targets metadata. The attacker goal
  is to cause clients to refuse to update the metadata later because the attacker's
  listed metadata version number (possibly MAX_INT) is greater than the new valid 
  version.  To recover from
```

JustinCappos: Would recommend something like this...

tuf-spec.md
```diff
@@ -1116,16 +1116,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+  fast-forward attacks after the repository has been compromised and recovered,
+  certain metadata files need to be deleted as specified in this section.
+  Please see [the Mercury
   paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
   for more details.
 
+    * **1.9.1**. **Targets recovery** If a threshold of targets keys have been
```
JustinCappos: You talk about top-level targets below, but not here.  Do you mean to differentiate them?  Does any delegated role's key loss result in the top-level targets file being deleted?

tuf-spec.md
```diff
@@ -1116,16 +1116,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+    removed in the new trusted root metadata compared to the previous trusted
+    root metadata, delete the old top-level targets and snapshot metadata
+    files.
```
JustinCappos: Why do we delete the snapshot metadata in this case?   If that key was compromised, won't it be rotated as well independently?

tuf-spec.md
```diff
@@ -1222,38 +1234,96 @@ non-volatile storage as FILENAME.EXT.
...
 
-    * **4.4.2**. Otherwise, recursively search the list of delegations in order
+    * **4.5.2**. Otherwise, recursively search the list of delegations in order
     of appearance.
 
-      * **4.4.2.1**. If the current delegation is a multi-role delegation,
+      * **4.5.2.1**. Let DELEGATE denote the current target role TARGETS is
+      delegating to.
+
+      * **4.5.2.2**. **Fast-forward attack recovery.** If a threshold of
```
JustinCappos: @mnm678 How does this relate to TAP 8?

tuf-spec.md
lukpueh: Good addition. Will add...

tuf-spec.md
lukpueh: What do you mean by "it will be rotated independently"?

If an attacker, who controls the top-level targets key, increases the version number in targets to  MAX_INT, the benign snapshot role might unwittingly adopt MAX_INT as the latest version number for targets. So we need to also delete snapshot after the compromised targets key has been rotated, otherwise we wouldn't get past verification step 3.3.3.

tuf-spec.md
lukpueh: I do mean top-level targets keys here too. I thought that might be clear, because root metadata only lists top-level targets keys.

Regarding your other question, why would we delete the top-level targets file if any delegated role's key is lost?  There is a detailed explanation about ffwd-recovery when delegated targets keys have been compromised in the commit message of https://github.com/theupdateframework/specification/pull/86/commits/f50f522b86e92f3f827e7af647ec67df4ac41ce1. Does that answer your question?

lukpueh:
Thanks for your reviews @JustinCappos, @joshuagl and @mnm678. I committed your suggestions and tried to answer the questions. When preparing this PR, I was a bit torn about how much background info and explanations should be part of the instructions. In order to not bloat the spec, I kept it to a minimum and instead wrote very detailed commit messages. Let me know if you think we should move some of that info from the commit messages directly into the spec.

tuf-spec.md
mnm678: TAP 8 adds a new way to rotate keys (rather than revoking and replacing), so if a threshold of keys are rotated under TAP 8 the same fast-forward recovery should happen.

tuf-spec.md
JustinCappos: I would think you could just rotate snapshot in this case too.  Or would this not work for some reason?

lukpueh: You are right, the consequences in regards to ffwd-recovery would be the same (through 1.9.2.), but with the overhead of rotating an "uncompromised" key.

If we do that, we should require, or at least strongly recommend, to always rotate snapshot, if targets is compromised.  And we should also generalize this for all roles. That is, we should specify:
- If top-level targets is rotated, also rotate snapshot (as repo) and delete trusted top-level targets (as client)
- If snapshot is rotated, also rotate timestamp (as repo) and delete trusted snapshot (as client) 
- If timestamp is rotated, delete trusted timestamp (as client)

These keys are all rotated via root, so the overhead seems feasible. But for consistency, we should do the same for delegated targets, i.e. we'd need to specify:
- If a delegated targets key is rotated, also rotate snapshot (as repo) and delete trusted delegated targets (as client)

Now this would mean that rotation of delegated targets keys in a delegating targets file, would also require rotating the snapshot keys in the root file, which doesn't seem desirable.

JustinCappos: I'm not sure why all of these rotations are needed.  Isn't this only true the case of a fast-forward attack?  Why in general do you need to rotate the snapshot when there hasn't been a FF attack on the targets role?

Also, if there is a FF attack on a delegated key, I think you don't need to rotate snapshot.  The key that was used to sign will not be trusted anymore (post "rotation") so delegations will end up pointing at a new delegated target.  

I'd also like to say we're not talking about rotation in a TAP 8 sense here.  TAP 8 has another mechanism for dealing with these sorts of issues.

tuf-spec.md
JustinCappos: Are you saying that the snapshot metadata file should be deleted whenever someone rotates a key?

tuf-spec.md
lukpueh: > I'm not sure why all of these rotations are needed. Isn't this only true the case of a fast-forward attack? Why in general do you need to rotate the snapshot when there hasn't been a FF attack on the targets role?

Yes, they are indeed only needed in case there was an FFWD attack, so we could add that extra bit to the recommendation. But does the repo know for sure that there was or wasn't an ffwd-attack? To me it seems safer to just assume it, so that clients can always resume updates. 

> Also, if there is a FF attack on a delegated key, I think you don't need to rotate snapshot. The key that was used to sign will not be trusted anymore (post "rotation") so delegations will end up pointing at a new delegated target.

Let's say an attacker compromises the delegated target `django` and bumps its version number in 'django.json' from `n` to `max_int`. The `snapshot` role  learns that 'django.json' is at `max_int` and, with good intentions,  writes `'django.json' : {'version': max_int}` to 'snapshot.json'.  Now the repo maintainer realize that  `django`'s key was compromised, replace it via its delegating role, and reset the version number to `n+1` in 'django.json', which `snapshot` adopts in 'snapshot.json'. Now since the `snapshot` key was not replaced, because it was not compromised, the client will not remove the trusted 'snapshot.json' that lists `max_int`, and thus will refuse to update to the lower `n+1` listed in the new 'snapshot.json' due to the rollback check in [3.3.3](https://github.com/theupdateframework/specification/blame/0f56aee998db130e6e1c43c5c7141217436b6d33/tuf-spec.md#L1188-L1195) (before this PR) or [4.5.2.3](https://github.com/theupdateframework/specification/blame/132f96a5c0d56f8b09c3bb775fa0c848298b2ba1/tuf-spec.md#L1274-L1278) (after this PR).

Or is it safe to assume that an attacker can't trick another non-compromised role into signing a wrong version number?


> I'd also like to say we're not talking about rotation in a TAP 8 sense here. TAP 8 has another mechanism for dealing with these sorts of issues.

Correct. By "rotation" above, I mean "a threshold of keys is removed". This PR also uses the latter wording instead of the former, in order to clarify this.

JustinCappos: Okay, after talking with @lukpueh , I think we're on the same page.  I was misremembering that something like the keyid is listed in the file name for a delegated targets role.  We do agree that cases where you know a fast forward attack did not occur (such as situations where a repository compromise did not happen) do not need to have the snapshot, etc. metadata deleted.

tuf-spec.md
```diff
@@ -1185,14 +1201,13 @@ non-volatile storage as FILENAME.EXT.
...
-    snapshot metadata file. Furthermore, any targets metadata filename that was
-    listed in the trusted snapshot metadata file, if any, MUST continue to be
-    listed in the new snapshot metadata file.  If any of these conditions are
-    not met, discard the new snaphot metadadata file, abort the update cycle,
-    and report the failure.
+    * **3.3.3**. The version number of the top-level targets metadata file, in
+    the trusted snapshot metadata file, if any, MUST be less than or equal to
+    its version number in the new snapshot metadata file. Furthermore, any
+    targets metadata filename that was listed in the trusted snapshot metadata
+    file, if any, MUST continue to be listed in the new snapshot metadata file.
```
trishankatdatadog: Hold on, why are we not checking that every version number for every targets metadata file (including delegations) in the previous snapshot, if any, is <= the current metadata file?

tuf-spec.md
```diff
@@ -1222,38 +1237,96 @@ non-volatile storage as FILENAME.EXT.
...
     step 5 (so that attackers cannot cause the client to waste excessive
     bandwidth or time).  Otherwise, if this role contains metadata about the
     desired target, then go to step 5.
 
-    * **4.4.2**. Otherwise, recursively search the list of delegations in order
+    * **4.5.2**. Otherwise, recursively search the list of delegations in order
     of appearance.
 
-      * **4.4.2.1**. If the current delegation is a multi-role delegation,
+      * **4.5.2.1**. Let DELEGATE denote the current target role TARGETS is
```
trishankatdatadog: Could we please say `DELEGATEE` instead of `DELEGATE`? Maybe ask @jhdalek55?

jhdalek55: I'm inclined to agree with Trishank here. While grammatically you might make a case that delegate is correct here, a delegate, when used as a noun, generally implies a person. "Delegatee" is defined as "someone or something to which something is delegated." In this case, when the responsibility for signing is being granted to a different entity, I feel like "delegatee" more clearly conveys this concept.

tuf-spec.md
lukpueh: It's described in detail in the corresponding commit message (f50f522b86e92f3f827e7af647ec67df4ac41ce1). 

The TLDR is, at this point in the process we don't know if any delegated targets keys were removed and we thus need to skip the rollback check.

tuf-spec.md
lukpueh: Done in 8070bdb

tuf-spec.md
trishankatdatadog: Hmmm, this basically breaks the workflow described in the Mercury paper, unless we're doing it elsewhere.

mnm678: The check in 4.5.2.7 looks for rollback attacks on delegated targets. Does that cover the workflow from the Mercury paper?

trishankatdatadog: Doesn't look like it, because Mercury prescribes checking vernum of _every_ targets metadata file.

mnm678: Yes, but you get almost the same security properties because every _downloaded_ targets metadata file is checked in 4.5.2.7. When the check was done on every targets metadata file it made it impossible to recover from a fast forward attack on a delegated targets metadata file.

trishankatdatadog: Hold on, I'm confused.

When you use `root` to rotate `timestamp` and `snapshot`, you discard the previous snapshot metadata file, no? And isn't rotating `root` the first thing we do? So why can't you check vernum for every targets metadata file?

JustinCappos: Do you mean for the targets metadata files you **previously downloaded** or somehow checking for all targets files, even those that were not downloaded?

trishankatdatadog: All targets files. Isn't this the Mercury workflow for snapshot?

mnm678: @trishankatdatadog If a fast forward attack occurs in a delegated targets file, does root still rotate timestamp and snapshot? This may require coordination between an independent developer who owns the delegated target and the repository manager which could slow the recovery time. If the root still rotates, then the check in 4.5.2.7 could be moved back up here.

trishankatdatadog: I see. Hmm, this requires careful thinking and modeling. I want to make sure we cover all of our bases. Right now, it doesn't look like it, but maybe I'm wrong.

lukpueh: Thanks for helping to clarify, @mnm678!

> I see. Hmm, this requires careful thinking and modeling. I want to make sure we cover all of our bases. Right now, it doesn't look like it, but maybe I'm wrong.

Can you elaborate, @trishankatdatadog? Why does it not look like we cover all of our bases? Maybe we should discuss this in person?

trishankatdatadog: Yes, I think we need a meeting.

trishankatdatadog:
@lukpueh Oh, right, this was the big PR that needed a meeting. Yes, could someone make this happen, please? But it can wait after TAP 5 @mnm678

lukpueh:
What stops us from landing this PR? cc @mnm678, @trishankatdatadog, @joshuagl, @JustinCappos.

mnm678:
> What stops us from landing this PR? cc @mnm678, @trishankatdatadog, @joshuagl, @JustinCappos.

I believe @trishankatdatadog wanted to meet about how this affects the Mercury workflow. I can schedule something for next week if that sounds alright.

trishankatdatadog:
> I believe @trishankatdatadog wanted to meet about how this affects the Mercury workflow. I can schedule something for next week if that sounds alright.

Yes please

mnm678:
For anyone interested in attending, please indicate your availability here: https://doodle.com/poll/hkfze8hwnmhvaq2p

mnm678:
The meeting is scheduled for Tuesday at 11am eastern time. Let me know if you would like an invite.

lukpueh:
Here's a recap of the meeting earlier this week, where we discussed this PR and the related issues with and without it.
Thanks for your time to discuss this, @trishankatdatadog, @mnm678, @joshuagl!

**Rollback prevention and fast-forward recovery prior to this PR**

In order to prevent rollback attacks the client checks that the version number of all targets metadata files (including **delegated targets**) in the trusted snapshot metadata file, **if any**, is less than or equal to their version number in the new snapshot metadata file.

In order to recover from a fast-forward attack (ffwd), the client temporarily disables that check after a targets role key rotation, by removing the trusted snapshot metadata file (see "if any" above).

**The issue**

For delegated targets, a client can only learn about a key rotation after having downloaded the delegating metadata, which happens after the snapshot versions check in the client workflow. As a consequence there is no ffwd recovery.

**The solution proposed in this PR**

Postpone the snapshot versions check for delegated targets until after their delegating targets were downloaded, so that the trusted snapshot can be removed upon a key rotation in time.

**The problem with this solution**

Rollback attacks go undetected for delegated targets that are not downloaded at the time when the rollback attack is visible in snapshot. And at a later time when the delegated targets is downloaded and the check would be performed, it is too late, because the rolled-back version in snapshot has already become the trusted version.

Usually, a second rollback check that compares the version of the delegated targets role with a prior trusted version of itself would solve this, but this does not work if it is downloaded for the first time.


**Solutions** (discussed in the meeting)

- Don't re-use a delegation role (name) if it is compromised, create a new role instead.

  **Con:** This is not ideal for hashed bin delegation, where role names are created according to a convention (e.g. `'00-07'`, `'08-0f'`, etc.). While we could use a different prefix (see related [python-tuf discussion](https://github.com/theupdateframework/tuf/pull/1106#issuecomment-681866823))  for a new role, it seems a bit cumbersome for a delegator to keep track of  different prefixes. Furthermore, the newly proposed [succinct hash bin delegation TAP](https://github.com/theupdateframework/tuf/pull/1106) only envisions one prefix for all bin roles.

- Require snapshot key rotation when a delegation is compromised.

  **Con:** This defeats the purpose of separating delegated role keys from top-level role keys.

**Another solution + variants**  (not yet discussed)

Perform snapshot versions check for all targets roles in snapshot before downloading any targets metadata (as prior to this PR), and again after the delegating role was downloaded (as per this PR). Then, if rollbacks are detected for delegated targets roles in the first check, don't abort the update, instead continue and:

(a) immediately resolve the delegation tree until all delegating roles of rolled-back roles have been downloaded, or
(b) keep a separate record of the rollbacks, until the corresponding delegating roles have been downloaded (possibly in later updates),

both to see if a key rotation for the rolled-back role took place, and the rollback might have been an attempt to recover from an ffwd.

**Con**
(a): requires additional state keeping across updates making the client more complex.
(b): might require to resolve the entire delegation tree.

lukpueh:
There's another problem with rollback check vs. FFWD recovery for delegated roles.

Even if we wait with the rollback check until the delegating role of a rolled-back role has been downloaded, we might be downloading the delegating role for the first time too and thus wouldn't detect a rotation. :(

trishankatdatadog:
Thanks for documenting our discussion, @lukpueh! I really do think choosing the simplest option and documenting the reasons why and its implications if the way to go. What do others think?

JustinCappos:
I feel like I'm missing a little context here (apologies about this).  Is
this worried about primarily when the key changes but the role name stays
the same?

I think the assumption when designing the ffwd defenses in Mercury was that
these attacks will be rare (assuming we have a defense), so using the
snapshot key is fine even if it is to recover from ffwd for a delegated
role.

lukpueh:
@JustinCappos and I just had a chat about this. He agrees that both [solutions from above](https://github.com/theupdateframework/specification/pull/86#issuecomment-687076718) make sense despite their drawbacks. The solutions are:
- don't reuse a compromised delegated targets role name, and/or
- rotate snapshot too when a delegated targets key is compromised

@JustinCappos also pointed out that the second rollback check, as removed in #65 and re-added in this PR (0ecf980), is indeed not necessary. The argument for the second rollback check was redundancy, so that an attacker needs to compromise snapshot and targets for a rollback attack, but in reality with the targets metadata >= version check in snapshot, even a compromised snapshot can't roll back.

@mnm678, @trishankatdatadog, what do you think? ... I'm starting to loose track here... :/

joshuagl:
> @JustinCappos and I just had a chat about this. He agrees that both [solutions from above](https://github.com/theupdateframework/specification/pull/86#issuecomment-687076718) make sense despite their drawbacks. The solutions are:
> 
> * don't reuse a compromised delegated targets role name, and/or

Would the onus be on the repository administrator to not reuse names? Or would we expect implementations to add some state-tracking for compromised names that can't be reused?

> * rotate snapshot too when a delegated targets key is compromised

Given that the above either requires diligent repository administrators or additional state tracking in implementations, I'm favouring this as the simplest solution.

trishankatdatadog:
> * don't reuse a compromised delegated targets role name, and/or

This looks the simplest to me, no need to bother repo admins, just the original delegator to the revoked delegatee, which may be a less onerous requirement. Bothering admins to reset snapshot should be last resort (as it can potentially affect everyone, and has bandwidth costs).

mnm678:
> > * rotate snapshot too when a delegated targets key is compromised
> 
> Given that the above either requires diligent repository administrators or additional state tracking in implementations, I'd favour this simpler solution.

I agree. This does require communicating with the repository maintainer, but it looks like the other viable solutions also require some additional communication. Also, as @lukpueh pointed out, this preserves rollback protection for the first time a delegated target is downloaded.

JustinCappos:
> Would the onus be on the repository administrator to not reuse names? Or
> would we expect implementations to add some state-tracking for compromised
> names that can't be reused?
>
It would be up to whomever is creating the new role.  Either the repo admin
or the user whose account was compromised.

Note also, that this only really matters if the key was not compromised but
somehow a fast forward account was done on the account.  (Maybe a runaway
script?)

lukpueh:
@JustinCappos:

> It would be up to whomever is creating the new role.  Either the repo admin
or the user whose account was compromised.

I agree. Plus coordination with the delegator is also required, because they need to add the new role to the delegating metadata, right?

> Note also, that this only really matters if the key was not compromised but
somehow a fast forward account was done on the account.  (Maybe a runaway
script?)

Why does it only really matter if the key was not compromised?

lukpueh:
> @JustinCappos also pointed out that the second rollback check, as removed in #65 and re-added in this PR (0ecf980), is indeed not necessary. The argument for the second rollback check was redundancy, so that an attacker needs to compromise snapshot and targets for a rollback attack, but in reality with the targets metadata >= version check in snapshot, even a compromised snapshot can't roll back.

@mnm678, @trishankatdatadog, @joshuagl, any thoughts? I submitted #65, but then let myself convince to revert it (see discussion https://github.com/theupdateframework/specification/issues/71#issuecomment-567230172 pp.). Looking at it now I think that I maybe misunderstood @mnm678's there. Rollback protection via snapshot seems to be effective even if snapshot is compromised.

I'd appreciate your opinions!

JustinCappos:
On Tue, Sep 8, 2020 at 4:44 PM lukpueh <notifications@github.com> wrote:

> @JustinCappos <https://github.com/JustinCappos>:
>
> It would be up to whomever is creating the new role. Either the repo admin
> or the user whose account was compromised.
>
> I agree. Plus coordination with the delegator is also required, because
> they need to add the new role to the delegating metadata, right?
>

Yes, for sure!

> Note also, that this only really matters if the key was not compromised but
> somehow a fast forward account was done on the account. (Maybe a runaway
> script?)
>
> Why does it only really matter if the key was not compromised?
>

If the key is compromised, you must change the key.  There isn't an option
to decide what to do and possibly just use the snapshot role to recover
from the fast forward attack.

>

lukpueh:
> If the key is compromised, you must change the key.  There isn't an option
to decide what to do and possibly just use the snapshot role to recover
from the fast forward attack.

Yes, you must change the key in the delegating role. But that alone won't allow a client to recover from a potential ffwd attack. So you do have to choose one of the above discussed options.

mnm678:
> Rollback protection via snapshot seems to be effective even if snapshot is compromised.

This is true if the rollback is to a version before the client last updated. If the client goes a week without updating, during which time there are 5 different releases, the client is only protected against a rollback to a version that is more than a week old. That being said, I think that this is an inherent tradeoff in the way snapshot handles rollback protection: the more often a client updates the better the rollback protection.

Looking back on the discussion in #71, I think my main concern was that rollback protection is not left to the timestamp role. The rollback protection using snapshot should be sufficient. Revoking the snapshot key for fast-forward attack recovery requires the support of root, and so would not introduce a rollback attack unless root is compromised.

trishankatdatadog:
What about the option of rotating the delegatee rolename? It requires cleaning up the revoked delegatee rolename later in the snapshot, but doesn't require immediate action from repo admins.

mnm678:
The main downside there would be hashed bin delegations or other places repositories generate rolenames.

trishankatdatadog:
> The main downside there would be hashed bin delegations or other places repositories generate rolenames.

I see. Yes, we should clarify which solution fits between in which case (if you control timestamp, snapshot, and the bins, then rotate snapshot; otherwise, just rotate delegatees for now, add revocation to snapshot, and wait for it). WDYT?

lukpueh:
@mnm678, do you have spare cycles to revive this work? I currently can't make it a priority, but it would be quite nice if we could land this PR in one way or another.

mnm678:
> @mnm678, do you have spare cycles to revive this work? I currently can't make it a priority, but it would be quite nice if we could land this PR in one way or another.

Sure!

mnm678:
Superseded by #150

PR #87, opened 2020-02-04
Add specification versioning- and release management instructions and checks

### LINKED ISSUE #73: Version release strategy for TUF...
We would like to make it possible to make changes to the TUF specification in a way that:

1. makes it easy to timestamp and archive versions of the TUF specification in semver style

1. enables there to be working drafts of new TUF versions.  In particular, changes that are backwards incompatible (major), backwards compatible (minor), and typo fixes (patch) in parallel.  

1. the manual effort to keep everything consistent is minimized.

To achieve these goals, I propose that there are three different "living" documents for the TUF spec (major-draft, minor-draft, and latest) that are edited in a very specific way.  To provide a conceptual model that is easy to understand, the documents form a hierarchy, major (backwards-incompatible) -> minor (backwards compatible) -> patch (typo fixes).  To minimize manual effort, changes in the document are only made to the drafts that need to be changed.  To achieve this, all changes to lower documents propagate to higher documents.  For example, if a typo is fixed in a patch, both the minor and major draft versions are updated to fix the typo.  If a backwards incompatible change is made to the specification, only the major document changes.  

Using such a hierarchy, it requires little effort to apply the correct semver number to a new release of TUF.  Simply take the prior version and increment the version number that applies to the appropriate patch, while zeroing out all lower version numbers.  For example, when at version 2.4.3 of the specification and accepting a minor draft, the new version number is 2.5.0.  In all cases, the newly approved draft becomes the latest.  If the major draft is accepted, then the previous major draft also becomes the new minor-draft and latest.

However, this does not handle the question of when to release a new version of the specification.  A new version of the specification is released in two scenarios.  First, whenever a patch change is accepted, the specification number is released.  Second, when there is consensus in the community that a specific minor version or major version is ready, this is accepted and finalized.

lukpueh:
Fixes #73 

- Add gitflow-based version- and release management instructions to the Versioning section of the README.

- Add links to "latest stable", "current draft", "new changes since latest stable" and "release history" of the specification in the header of the README document.
(Note that the version number of the latest stable version of the specification is no longer shown in the README, which eliminates the need to update the README with every new release.)

- Add script to run on Travis upon a GitHub pull request, that checks that versioning- and release management rules are followed, and configure travis accordingly.

Reviewers may focus on the changes in `README.rst`. 
The `check_release.py`-script and `.travis.yml`-configuration are only there to help us follow the version/release strategy.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,10 @@
 The Update Framework specification
 ----------------------------------
 
-Latest: `version 1.0.0 <https://github.com/theupdateframework/specification/blob/master/tuf-spec.md>`_
+- `latest stable <https://github.com/theupdateframework/specification/blob/master/tuf-spec.md>`_
+- `current draft <https://github.com/theupdateframework/specification/blob/draft/tuf-spec.md>`_
+- `new changes since latest stable <https://github.com/theupdateframework/specification/compare/master..draft>`_
+- `release history <https://github.com/theupdateframework/specification/tags>`_
 
 
 Contact
@@ -31,7 +34,33 @@ Versioning
 ----------
 
 The TUF specification uses `Semantic Versioning 2.0.0 <https://semver.org/>`_
-for its version numbers.
+(semver) for its version numbers, and a gitflow-based release management:
+
+- The 'master' branch of this repository always points to the latest stable
+  version of the specification.
+- The 'draft' branch of this repository always points to the latest development
+  version of the specification and must always be based off of the latest
+  'master' branch.
+- Contributors must submit changes as pull requests against these branches,
+  depending on the type of the change (see semver rules).
+- For patch-type changes, pull requests may be submitted directly against the
+  'master' branch.
+- For major- and minor-type changes, pull requests must be submitted against
+  the 'draft' branch.
+- Maintainers may, from time to time, decide that the 'draft' branch is ready
+  for a new major or minor release, and submit a pull request from 'draft'
+  against 'master'.
+- Before merging a branch with 'master' the 'last modified date' and 'version'
+  in the specification header must be bumped.
+- Merges with 'master' that originate from the 'draft' branch must bump either
+  the major or minor version number.
+- Merges with 'master' that originate from any other branch must bump the patch
+  version number.
+- Merges with 'master' must be followed by a git tag for the new version
+  number.
+- Merges with 'master' must be followed by a rebase of 'draft' onto 'master'.
+
+
 
 Acknowledgements
 ----------------
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: python
+
+# Disable auto-cloning and ...
+git:
+  clone: false
+
+# ... instead manually fetch and checkout the pull request source branch, which
+# is expected by check_release.py
+# https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally
+install:
+  - git clone --depth=50 https://github.com/${TRAVIS_REPO_SLUG}.git ${TRAVIS_REPO_SLUG}
+  - cd ${TRAVIS_REPO_SLUG}
+  - git fetch origin pull/${TRAVIS_PULL_REQUEST}/head:${TRAVIS_PULL_REQUEST_BRANCH}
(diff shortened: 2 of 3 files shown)


lukpueh:
I just configured branch restrictions for the `master` branch to:
- require one review approval before merging
- dismiss prior approvals when new commits are pushed
- require status check to pass before merge, i.e. `check_release.py`
- require  branches to be up to date before merging

Repo administrator can still bypass any of these restrictions.

lukpueh:
IMHO the next steps should be:
1. Review and merge this PR (despite failing travis check)
1. Create and merge a PR that removes TAP 3 and TAP 4 from the spec, i.e. removes 2 lines in section 1.5.4. (despite failing travis check)
   1. (Optionally merge in  #86 (despite failing travis check))
1. Make a formal 1.0.0 spec release, i.e. `git tag v1.0.0`
1. Henceforth require folks to follow the release/versioning strategy suggested by this PR and travis check to pass

After that the next steps will be to add TAP 4 and release 1.1.0, and add TAP 3 (#57) and release 2.0.0.

README.rst
```diff
@@ -31,7 +34,33 @@ Versioning
...
+  for a new major or minor release, and submit a pull request from 'draft'
+  against 'master'.
+- Before merging a branch with 'master' the 'last modified date' and 'version'
+  in the specification header must be bumped.
+- Merges with 'master' that originate from the 'draft' branch must bump either
+  the major or minor version number.
+- Merges with 'master' that originate from any other branch must bump the patch
+  version number.
+- Merges with 'master' must be followed by a git tag for the new version
+  number.
```
lukpueh: This is currently not verified by `check_release.py`, because the script is executed before the merge, and the tag should probably be created after the merge.  It would be nice to automate tag or release creation.

lukpueh: Any thoughts, anyone?

JustinCappos: I agree this would be nice to automate.

README.rst
```diff
@@ -31,7 +34,33 @@ Versioning
...
+  version of the specification.
+- The 'draft' branch of this repository always points to the latest development
+  version of the specification and must always be based off of the latest
+  'master' branch.
+- Contributors must submit changes as pull requests against these branches,
+  depending on the type of the change (see semver rules).
+- For patch-type changes, pull requests may be submitted directly against the
+  'master' branch.
+- For major- and minor-type changes, pull requests must be submitted against
+  the 'draft' branch.
```
JustinCappos: It's slightly odd to me that the person sending a PR doesn't indicate minor vs major.  Also, it sounds like those are getting intertwined in a way we may not want.  What if we want to push a release with some backwards compat (minor version) changes but not others that are not backwards compatible?  Do we disentangle these changes at that time?

JustinCappos: I'd prefer to have a model where there are major, minor, and patch branches and they interrelate in the way your master and draft branches do.

I like the overall approach though!

README.rst
lukpueh: Okay, I can look into it.

README.rst
lukpueh: > It's slightly odd to me that the person sending a PR doesn't indicate minor vs major. 

I think the person who sends the PR should not be bothered with the question into what future release the submitted change goes, regardless of it being major, minor or patch. This is the task of the maintainer. Maybe I should add a note to this section that says maintainers may [change the target of a pending PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request)?

> Also, it sounds like those are getting intertwined in a way we may not want. What if we want to push a release with some backwards compat (minor version) changes but not others that are not backwards compatible? Do we disentangle these changes at that time?

Coordinating major and minor releases can still be done, either by holding back PRs from being merged into `draft`, or, if it's too late for that, by shuffling around the git history on `draft`. The consensus in #73 seemed to be that the operational simplicity of having one working branch, plus the option to directly patch `master`, outweighs the flexibility of multiple permanent draft branches.

> I like the overall approach though!

Thanks.

mnm678: I agree with @lukpueh, we want to make PRs as simple as possible. As long as we merge draft into master regularly, I don't think that we will need any additional branches.

PR #89, opened 2020-02-05
Use consistent URLs for the TUF home page

joshuagl:
Readers were referred to two different URLs for the TUF home page, both
of which redirect to theupdateframework.io -- therefore use only this
URL consistently throughout.

Signed-off-by: Joshua Lock <jlock@vmware.com>

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -54,7 +54,7 @@ repo](https://github.com/theupdateframework/specification/issues).
    Work on TUF began in late 2009.  The core ideas are based off of previous
    work done by Justin Cappos and Justin Samuel that identified security flaws
    in all popular Linux package managers.  More information and current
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks @joshuagl. We're currently in the process of updating the DNS entries for the existing TUF domains (there are several). It should soon be possible to use 'www.theupdateframework.com' without redirects. I think we don't need to change this in the spec.

joshuagl:
Ah, that makes sense. In which case we probably only need to update the reference to www.updateframework.com -> www.theupdateframework.com

lukpueh:
We should soon be able to use 'theupdateframework.org', 'theupdateframework.com' and 'updateframework.com', with and without www. :) I don't think we need to change it.

joshuagl:
Understood. I'll close the PR, thanks @lukpueh

Issue #38, opened 2019-03-24
Does timestamp.json still require `length` and `hashes`?


erickt:
In 4.6 of the spec, the timestamp role METAFILES is described as:

> METAFILES is the same is described for the snapshot.json file. In the case of the timestamp.json file, this will commonly only include a description of the snapshot.json file.

Where the METAFILES described in snapshot.json is:

```
METAFILES is an object whose format is the following:

 { METAPATH : {
       "version" : VERSION }
   , ...
 }
METAPATH is the metadata file's path on the repository relative to the metadata base URL.

VERSION is listed for the root file and all other roles available on the repository.
```

However, the example in the spec includes the `size` and `hashes` field:

```
...
 "signed": {
  "_type": "timestamp",
  "spec_version": "1",
  "expires": "2030-01-01T00:00:00Z",
  "meta": {
   "snapshot.json": {
    "hashes": {
     "sha256": "c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681"
    },
    "length": 1007,
    "version": 1
   }
...
```

Are the `length` and `hashes` field optional for the timestamp.json role, or are they required? I'm not sure if the arguments made in https://github.com/theupdateframework/tuf/issues/451 and https://github.com/theupdateframework/tuf/issues/465 to make these fields optional for the snapshot role also apply to the timestamp role.

trishankatdatadog:
Since `timestamp.json` is always constant size in practice (but even if attackers inflate the file, we have upper bounds on downloading files in practice), this is not a problem. It would be appreciated if you could send a PR clarifying this :)

awwad:
As to Eric's question about the hashes:

The TUF specification is currently written to require that timestamp list hashes for snapshot, yes; this is explicit in 2.1.4 in the spec.

I cannot now think of a reason that the hash of snapshot in timestamp (instead of just the version of snapshot in timestamp) is really that helpful.  I may be forgetting something.  Anyone else?

Issue #28, opened 2018-08-30
It is probably not helpful for Timestamp to list Snapshot's hash


awwad:
Timestamp and Root are likely to be the two most-downloaded files on a repository. Currently, Timestamp lists information for Snapshot in a different way than Snapshot lists information for Targets and delegated targets files: timestamp lists the hash of snapshot in addition to the version number of snapshot. It is not clear that this is actually useful.

The costs to this are a few:
- It adds size to the timestamp role file (one hash where timestamp would otherwise not contain hashes)
- It makes the timestamp role file a little harder to read and understand.
- It makes conceptually reconciling timestamp and snapshot definitions a bit harder for implementers / new folks.
- It requires a bulkier and less intuitive programmatic representation (in formats.py and tuf_metadata_definitions.asn1).

[Snapshot definition](https://github.com/theupdateframework/specification/blame/f5a6f00a2a5d8c4b05e24bbe5de718aa09da167c/tuf-spec.md#L690-L711)
[Timestamp definition](https://github.com/theupdateframework/specification/blame/f5a6f00a2a5d8c4b05e24bbe5de718aa09da167c/tuf-spec.md#L905-L953)

PR #90, opened 2020-02-05
 Clarify timestamp.json METAFILES format

joshuagl:
A read-through of the spec made me question the definition of METAFILES for `snapshot.json`.
In step 3 of the client application detailed workflow the spec states that the client download 
> "up to the number of bytes specified in the timestamp metadata file"

yet the description of METAFILES for `snapshot.json` suggests the object behaves the same as METAFILES in `timestamp.json` where LENGTH is optional.

As @erickt raised a similar concern in #38 here's a pull request to try and clarify.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -975,9 +975,30 @@ repo](https://github.com/theupdateframework/specification/issues).
          "meta" : METAFILES
        }
 
-   METAFILES is the same as is described for the snapshot.json file.  In the
-   case of the timestamp.json file, this MUST only include a description of the
-   snapshot.json file and the LENGTH of the snapshot.json file is REQUIRED.
+   METAFILES is an object whose format is the following:
+
+       { METAPATH : {
+             "version" : VERSION,
+             "length" : LENGTH, |
+             "hashes" : HASHES }
+         , ...
+       }
+
+   METAPATH is the snapshot.json file's path on the repository relative to the
+   metadata base URL.
+
+   VERSION is the version of the snapshot metadata as listed in the
+   snapshot.json file.
+
+   LENGTH is the integer length in bytes of the snapshot metadata file.
+
+   HASHES is the dictionary that specifies one or more hashes of the snapshot
+   metadata file, including the cryptographic hash function.  For example:
+   { "sha256": HASH, ... }.
+
+   Note: METAFILES is the same as is described for the snapshot.json file, only
+   the LENGTH and HASHES are not optional and in the case of the timestamp.json
+   file, this MUST only include a description of the snapshot.json file.
 
    A signed timestamp.json example file:
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1165,10 +1165,10 @@ the timestamp metadata file.  In either case, the client MUST write the file to
 non-volatile storage as FILENAME.EXT.
 
   * **3.1**. **Check against timestamp metadata.** The hashes and version
-  number of the new snapshot metadata file MUST match the hashes and version
-  number listed in timestamp metadata.  If hashes and version do not match,
-  discard the new snapshot metadata, abort the update cycle, and report the
-  failure.
+  number of the new snapshot metadata file MUST match the hashes (if any) and
+  version number listed in the trusted timestamp metadata.  If hashes and
+  version do not match, discard the new snapshot metadata, abort the update
+  cycle, and report the failure.
 
   * **3.2**. **Check signatures.** The new snapshot metadata file MUST have
   been signed by a threshold of keys specified in the trusted root metadata
@@ -1213,11 +1213,12 @@ VERSION_NUMBER is the version number of the targets metadata file listed in the
 snapshot metadata file.  In either case, the client MUST write the file to
 non-volatile storage as FILENAME.EXT.
 
-  * **4.1**. **Check against snapshot metadata.** The hashes (if any), and
-  version number of the new targets metadata file MUST match the trusted
-  snapshot metadata.  This is done, in part, to prevent a mix-and-match attack
-  by man-in-the-middle attackers.  If the new targets metadata file does not
-  match, discard it, abort the update cycle, and report the failure.
+  * **4.1**. **Check against snapshot metadata.** The hashes and version
+  number of the new targets metadata file MUST match the hashes (if any) and
+  version number listed in the trusted snapshot metadata.  This is done, in
+  part, to prevent a mix-and-match attack by man-in-the-middle attackers.  If
+  the new targets metadata file does not match, discard it, abort the update
+  cycle, and report the failure.
 
   * **4.2**. **Check for an arbitrary software attack.** The new targets
   metadata file MUST have been signed by a threshold of keys specified in the
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -709,10 +709,9 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 * **4.4. File formats: snapshot.json**
 
-   The snapshot.json file is signed by the snapshot role.  It lists the version
-   numbers of only the top-level targets and all delegated targets role metadata.
-   The metadata length and hashes are OPTIONAL for the top-level targets and
-   all delegated targets roles.
+   The snapshot.json file is signed by the snapshot role. It MUST list the
+   version numbers of the top-level targets metadata and all delegated targets
+   metadata. It MAY also list their lengths and file hashes.
 
    The "signed" portion of snapshot.json is as follows:
 
@@ -732,18 +731,23 @@ repo](https://github.com/theupdateframework/specification/issues).
          , ...
        }
 
-   METAPATH is the metadata file's path on the repository relative to the
-   metadata base URL.
+   METAPATH is the file path of the metadata on the repository relative to the
+   metadata base URL. For snapshot.json, these are top-level targets metadata
+   and delegated targets metadata.
 
-   VERSION is listed for the top-level targets and all delegated targets roles
-   available on the repository.
+   VERSION is the integer version number as shown in the metadata file at
+   METAPATH.
 
-   LENGTH is the integer length in bytes of the metadata file. It is
-   OPTIONAL for all roles.
+   LENGTH is the integer length in bytes of the metadata file at METAPATH. It
+   is OPTIONAL and can be omitted to reduce the snapshot metadata file size. In
+   that case the client MUST use a custom download limit for the listed
+   metadata.
 
-   HASHES is the dictionary that specifies one or more hashes, including
-   the cryptographic hash function.  For example: { "sha256": HASH, ... }. It is
-   OPTIONAL for all roles.
+   HASHES is a dictionary that specifies one or more hashes of the metadata
+   file at METAPATH, including their cryptographic hash function. For example:
+   { "sha256": HASH, ... }. HASHES is OPTIONAL and can be omitted to reduce
+   the snapshot metadata file size.  In that case the repository MUST guarantee
+   that VERSION alone unambiguously identifies the metadata at METAPATH.
 
    A snapshot.json example file:
 
@@ -958,8 +962,8 @@ repo](https://github.com/theupdateframework/specification/issues).
 
 * **4.6. File formats: timestamp.json**
 
-   The timestamp file is signed by a timestamp key.  It indicates the
-   latest versions of other files and is frequently resigned to limit the
+   The timestamp file is signed by a timestamp key.  It indicates the latest
+   versions the snapshot metadata and is frequently resigned to limit the
    amount of time a client can be kept unaware of interference with obtaining
    updates.
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -2,7 +2,7 @@
 
 Last modified: **5 March 2020**
 
-Version: **1.0.0**
+Version: **1.0.1**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -975,9 +975,9 @@ repo](https://github.com/theupdateframework/specification/issues).
          "meta" : METAFILES
        }
 
-   METAFILES is the same is described for the snapshot.json file.  In the case
-   of the timestamp.json file, this MUST only include a description of the
-   snapshot.json file.
+   METAFILES is the same as is described for the snapshot.json file.  In the
+   case of the timestamp.json file, this MUST only include a description of the
+   snapshot.json file and the LENGTH of the snapshot.json file is REQUIRED.
 
    A signed timestamp.json example file:
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -975,30 +975,9 @@ repo](https://github.com/theupdateframework/specification/issues).
          "meta" : METAFILES
        }
 
-   METAFILES is an object whose format is the following:
-
-       { METAPATH : {
-             "version" : VERSION,
-             "length" : LENGTH, |
-             "hashes" : HASHES }
-         , ...
-       }
-
-   METAPATH is the snapshot.json file's path on the repository relative to the
-   metadata base URL.
-
-   VERSION is the version of the snapshot metadata as listed in the
+   METAFILES is the same is described for the snapshot.json file.  In the case
+   of the timestamp.json file, this MUST only include a description of the
    snapshot.json file.
-
-   LENGTH is the integer length in bytes of the snapshot metadata file.
-
-   HASHES is the dictionary that specifies one or more hashes of the snapshot
-   metadata file, including the cryptographic hash function.  For example:
-   { "sha256": HASH, ... }.
-
-   Note: METAFILES is the same as is described for the snapshot.json file, only
-   the LENGTH and HASHES are not optional and in the case of the timestamp.json
-   file, this MUST only include a description of the snapshot.json file.
 
    A signed timestamp.json example file:
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -963,7 +963,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 * **4.6. File formats: timestamp.json**
 
    The timestamp file is signed by a timestamp key.  It indicates the latest
-   versions the snapshot metadata and is frequently resigned to limit the
+   version of the snapshot metadata and is frequently resigned to limit the
    amount of time a client can be kept unaware of interference with obtaining
    updates.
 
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -979,9 +979,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
        }
 
-   METAFILES is the same is described for the snapshot.json file.  In the case
-   of the timestamp.json file, this MUST only include a description of the
-   snapshot.json file.
+   METAFILES is an object whose format is the following:
+
+       { METAPATH : {
+             "version" : VERSION,
+             ("length" : LENGTH, |
```
```suggestion
             "length" : LENGTH, |
```

tuf-spec.md
```diff
@@ -979,9 +979,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+              "hashes" : HASHES) }
```
```suggestion
              "hashes" : HASHES }
```

tuf-spec.md
```diff
@@ -979,9 +979,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+         , ...
+       }
+
+   METAPATH is the snapshot.json file's path on the repository relative to the
+   metadata base URL.
+
+   VERSION is the version the snapshot target as listed in the snapshot.json
+   file.
+
+   LENGTH is the integer length in bytes of the metadata file.
```
```suggestion
   LENGTH is the integer length in bytes of the snapshot metadata file.
```

tuf-spec.md
```diff
@@ -979,9 +979,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+
+   HASHES is the dictionary that specifies one or more hashes, including
```
```suggestion
   HASHES is the dictionary that specifies one or more hashes of the snapshot metadata file, including
```

joshuagl:
I've force pushed an update which includes @trishankatdatadog's suggests, thanks for the review!

tuf-spec.md
lukpueh: Is this different from other "or"-groups in the spec, where we do use parentheses (see [#L734-L735](https://github.com/theupdateframework/specification/blame/97312ddd239746a213694ce772e9bccbf4123f39/tuf-spec.md#L734-L735) and [#L840-L841](https://github.com/theupdateframework/specification/blame/97312ddd239746a213694ce772e9bccbf4123f39/tuf-spec.md#L840-L841))?

If not, I suggest we use a consistent notation, with or without parentheses.

lukpueh: Oh, I understand now. The lack of parentheses indicates that one of the values must be provided.

lukpueh: But if only one of them is mandatory, then the routine from the client workflow that @joshuagl quotes above , i.e. 
> "up to the number of bytes specified in the timestamp metadata file"

would fail in the case timestamp only lists  `hashes`.

joshuagl: Very good point. If we keep the METAPATH object as is with optional fields, then the client should fetch either `length` or an application specified size, and in the latter case compare to the `hash`.

Perhaps `length` should be mandatory here, the client workflow seems to imply that it is.

joshuagl:
Thanks for the review @lukpueh. You are right, with the change as-is we would need to rephrase the client workflow.

However, having read #28 and re-read #38 it does seem that we might be able to simplify the listing of `snapshot.json` in `timestamp.json`. 

If we remove the hashes, does it make sense to list `length` still? Or would it be better to allow the application to _"download up to Z number of bytes"_ as with other metadata file downloads?

lukpueh:
I can't see how the hashes and lengths fields depend on each other. So I don't think there would be a problem in keeping/mandating only one of them. But I do wonder if we even need to keep/mandate either of them.

I assume the main argument for `lengths` is to have an exact download upper bound. But as @trishankkarthik points out in #38, timestamp should always have pretty much the same size (modulo a few bytes more or less due to snapshot's version), so a *"download up to Z number of bytes"*-strategy should be good enough.

`hashes` OTOH seems to make the connection between timestamp and snapshot more robust. But if there is a unique mapping from timestamp to snapshot via snapshot's version number, `hashes` shouldn't be needed.  Note that snapshot's integrity is guaranteed by snapshot's signature. And the version number should indeed always be unique, in order to prevent rollback attacks (see https://github.com/secure-systems-lab/peps/issues/69 for related discussions).

lukpueh:
Ping @mnm678, @JustinCappos, @trishankatdatadog. Your opinions on the matter would be appreciated!  :)

We should be able to close both #28 and #38 with this PR. This can be done by removing/optionalizing hashes and lengths in timestamp (for the reasons presented above and in the issues), or by making an argument for not doing so.

Either way, clarifying the relevant section in the spec, as @joshuagl does, is a good idea.

trishankatdatadog:
I think we might as well make timestamp consistent with snapshot, so if the latter lists only version numbers, then that's what the former should do, too. I can't imagine we lose much security this way, especially since, in practice, timestamp and snapshot share the same online key anyway, and as long as we set an upper bound on the size of any metadata downloaded. So, length and hashes may go back to being optional, just like in snapshot.

mnm678:
I think that we should remove/make optional the hashes and lengths in timestamp to match the requirement in snapshot. As @trishankatdatadog  mentioned, the only benefit gained from the hashes is extra verification of the contents of snapshot, which is redundant as snapshot is alright signed by the snapshot key (and in practice this is often the same as the timestamp key). 

As for the question of whether to remove them or make them optional, I would propose we make them optional for backwards compatibility, and possibly remove them in a future major release.

JustinCappos:
I do feel more nervous about removing the hash and length from timestamp.  The length of snapshot could increase dramatically as new targets are added.  It could also decrease after a rotation of the snapshot key and clean up of outdated targets files.  

I'll list a (semi-contrived) situation where version number, hash, and length are all important to have.
1. If missing the length, but with a hash and version number.  An attacker can launch an endless data attack more easily without compromising any keys.  
2. If missing the hash, but with a length and version number.  An attacker which has stolen the timestamp and snapshot key and who can answer client requests (e.g., a compromised repo), can give different clients the same timestamp but different snapshot metadata.  This may allow bifurcation of client requests without doing things like rapidly increasing the version number or giving different timestamp files, both of which would be more noticeable.  (This is fairly contrived.  I'm not really sure the impact here, but it seems worrying.)
3. If missing the version number, but with a hash and a length.  An attacker that has compromised the timestamp only, may generate a timestamp file for a future time that lists an old snapshot file.  A client updating later, but before the future time, would not download the new snapshot file, thus causing a freeze attack without the attacker needing to maintain control of the repo.  (The repo could rotate timestamp to fix this.)

My off-the-cuff thoughts are: having a version number is good, having a length is good, having a hash is good, and having all three seems to be the best.  I believe it may be possible to drop one or more (but we would want to think very carefully about this), but is there a reason not to list all three?

lukpueh:
> 1. If missing the length, but with a hash and version number.  An attacker can launch an endless data attack more easily without compromising any keys.

I realize now that the constant size of timestamp as argument  against `lengths` in timestamp (in my [comment above](https://github.com/theupdateframework/specification/pull/90#issuecomment-582849329) and [a comment in #38](https://github.com/theupdateframework/specification/issues/38#issuecomment-476309010)) is nonsense, since the field specifies snapshot's size and not timestamp's. 🤦‍♂ Sorry for the confusion.

That means an upper bound for snapshot, either via `lengths` in timestamp, or defined by the client (as we do for other metadata) is necessary to prevent endless data attacks.

> 2. If missing the hash, but with a length and version number.  An attacker which has stolen the timestamp and snapshot key and who can answer client requests (e.g., a compromised repo), can give different clients the same timestamp but different snapshot metadata.  [...]

AFAICS, with the given attacker capabilities, they could achieve the same by serving different timestamp and different snapshot metadata to different clients, even if timestamp includes a hash of snapshot. Or is there an advantage for the attacker to serve only different snapshot metadata? 

> 3. If missing the version number, but with a hash and a length.  An attacker that has compromised the timestamp only, may generate a timestamp file for a future time that lists an old snapshot file.  [...]

I think we all agree that snapshot's version number in timestamp.json is important to prevent rollback attacks, despite a compromised timestamp role.

----

It seems like the removal/optionalization of `hashes` and `lengths` was motivated by bandwidth concerns in the Mercury paper (see https://github.com/heartsucker/rust-tuf/issues/104 and https://github.com/theupdateframework/tuf/issues/465). However, the paper does not seem to mention the `lengths` field (it only talks about `hashes` vs. `version`), nor the timestamp role (it only talks about those fields in snapshot).

To summarize my thoughts on those fields:
- `version` 
  - required to fail early on rollback attacks
  - alternative to `hashes` for mapping the referenced metadata. The mapping should be 1-1, but could become 1-n, e.g.  if version numbers are re-used or in @JustinCappos' bifurcation scenario (see 2. above). 
- `hashes`
  - more robust mapping of referenced metadata, compared to `version` (1-1 if no hash collisions)
  - strong impact on bandwidth
- `lengths` 
  - precise and flexible way of preventing endless-data attacks,
  - a client-defined upper bound could be a bandwidth-effective but less precise and less flexible alternative
 
And to summarize my thoughts on the affected roles:
- timestamp:
  - most downloaded role
  - only has one entry, i.e. snapshot.json
- snapshot: 
  - second most downloaded role
  - can have many entries , i.e. targets.json and all <delegated dargets>.json

Please correct me if I'm wrong or missing something!

trishankatdatadog:
> My off-the-cuff thoughts are: having a version number is good, having a length is good, having a hash is good, and having all three seems to be the best. I believe it may be possible to drop one or more (but we would want to think very carefully about this), but is there a reason not to list all three?

Yeah, but shouldn't the `snapshot` also list all three by that logic? Don't we accept some tradeoff in security for b/w performance there? Here there is no such demand, but I find all the scenarios above fairly contrived. Having said that, there's absolutely no reason to drop all 3 in `timestamp` except for some aesthetic consistency.

JustinCappos:
Snapshot doesn't have hashes because the size becomes quite large overall
and our Mercury work shows that the version number is even more valuable to
have than the hash in most cases.

For timestamp, it's a single entry so I think doesn't matter in the same
way...

lukpueh:
> Snapshot doesn't have hashes because the size becomes quite large overall and our Mercury work shows that the version number is even more valuable to have than the hash in most cases. For timestamp, it's a single entry so I think doesn't matter in the same way...

Interestingly, the specification currently suggests the opposite. For [timestamp.json it says](https://github.com/theupdateframework/specification/blame/97312ddd239746a213694ce772e9bccbf4123f39/tuf-spec.md#L970-L971):

> Timestamp files will potentially be downloaded very frequently.  Unnecessary
   information in them will be avoided.

lukpueh:
Based on the discussions here, I took the liberty to hijack this PR, reverting the commits and instead 
- adding a client-set download limit as fallback if snapshot length is not set in timestamp.json, and
- generally clarifying METAFILES and its fields for timestamp.json and snapshot.json

@joshuagl, I hope you don't mind. This is just a suggestion. We can easily reset to what it was before.

@trishankatdatadog, @JustinCappos, would you kindly take another look? `length` and `hashes` is now again optional in both timestamp.json and snapshot.json. And I added some context, why one would want to omit them (i.e. bandwidth), and under what circumstances they may be omitted (i.e. fallback download limit, no reused version numbers). 

We could further add a recommendation for when those fields should be omitted (i.e. for timestamp: `'low expiration'` and for snapshot:   `'low expiration' x 'many (delegated) targets'`). What do you think?

joshuagl:
I don't mind at all, this looks like a good set of clarifications to me.

I do like the idea of adding a recommendation on when optional fields could be omitted.

tuf-spec.md
```diff
@@ -1153,8 +1157,10 @@ as FILENAME.EXT.
   file.  If the new timestamp metadata file has expired, discard it, abort the
   update cycle, and report the potential freeze attack.
 
-**3**. **Download snapshot metadata file**, up to the number of bytes specified
-in the timestamp metadata file.  If consistent snapshots are not used (see
+**3**. **Download snapshot metadata file**, up to the number of bytes
+specified in the timestamp metadata file. If not specified, download up to a
+number of bytes set by the authors of the application using TUF. This may be
```
lukpueh: I purposefully say *"a number of bytes"* instead of *"VAR number of bytes"* because we already use *X* and *Y* above, and *Z* below, so I wasn't sure what VAR to use. If others want, we can make the wording consistent. Any thoughts?

joshuagl: I noticed that, we could shuffle the variables so that we use _W_, _X_, _Y_ and _Z_.

lukpueh:
> No brackets in METAPATH now, since the fields are no longer optional, please lose them

Which brackets are you referring to, @trishankatdatadog? The current diff of this PR doesn't change anything about the metadata format. Do you want to make `version`, `length` and `hashes` mandatory for *snapshot* and *timestamp*? 

How does that fit in with what you say [above](https://github.com/theupdateframework/specification/pull/90#issuecomment-583474704) that  *"we might as well make timestamp consistent with snapshot, so if the latter lists only version numbers, then that's what the former should do"*. 

Also, the specification [currently claims](https://github.com/theupdateframework/specification/blame/97312ddd239746a213694ce772e9bccbf4123f39/tuf-spec.md#L970-L971) *"Timestamp files will potentially be downloaded very frequently. Unnecessary information in them will be avoided."* Is that not a reference to the suggestions of the Mercury paper?

tuf-spec.md
lukpueh: Good idea. Let's wait till we know in what direction this PR should go. If we decide to make `length` mandatory, we don't need the client-defined upper bound.

trishankatdatadog:
@lukpueh Yeah, I think we should make a determination whether we're going to use them or not. Let's discuss in the next community meeting.

P.S. At the time that part of the spec was written, I don't think the Mercury paper was around.

lukpueh:
Thanks for the comment @trishankatdatadog. What changes do you request? In the current version of this PR the `METAFILES` data format is defined in section "4.4. File formats: snapshot.json", marking the `length` and `hashes` field as optional. In section "4.6. File formats: timestamp.json" , we say that `METAFILES` is *is the same is described for the snapshot.json file*, i.e. `length` and `hashes` are also flexible.

joshuagl:
Can we update step **3**. **Download snapshot metadata file** to use a variable (" so that it's consistent with the other steps?

lukpueh:
Good call, @joshuagl. Do you want to push a commit with the variables W,X,Y,Z you proposed above?

joshuagl:
> Good call, @joshuagl. Do you want to push a commit with the variables W,X,Y,Z you proposed above?

Done.

tuf-spec.md
```diff
@@ -736,18 +735,23 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+   metadata.
 
-   HASHES is the dictionary that specifies one or more hashes, including
-   the cryptographic hash function.  For example: { "sha256": HASH, ... }. It is
-   OPTIONAL for all roles.
+   HASHES is a dictionary that specifies one or more hashes of the metadata
+   file at METAPATH, including their cryptographic hash function. For example:
+   { "sha256": HASH, ... }. HASHES is OPTIONAL and can be omitted to reduce
+   the snapshot metadata file size.  In that case the repository MUST guarantee
+   that VERSION alone unambiguously identifies the metadata at METAPATH.
```
mnm678: We might want to note when length and hashes should be included (ie backwards compatibility for both, using length as an exact download size)

trishankatdatadog: Backwards compatibility is one reason, yes, but I believe the more important reason is to protect against certain security attacks. Please see Section 5.6 of [Mercury](https://www.usenix.org/system/files/conference/atc17/atc17-kuppusamy.pdf).

trishankatdatadog: Also, if we are bringing back hashes and length here, we should add the corresponding checks in the download process.

lukpueh: >We might want to note when length and hashes should be included

@mnm678, @trishankatdatadog: Good idea. Could you push a commit with such a note?

> Also, if we are bringing back hashes and length here, we should add the corresponding checks in the download process.

@trishankatdatadog: What's missing from the existing ([L1136-1138](https://github.com/theupdateframework/specification/pull/90/files#diff-94b2d37a08c587082c1ea1aaead6f973R1136-R1138)) and newly proposed ([L11160-1163](https://github.com/theupdateframework/specification/pull/90/files#diff-94b2d37a08c587082c1ea1aaead6f973R1160-R1163)) download checks?

trishankatdatadog: @lukpueh Those are good checks, but:

* 3.1 doesn't mention that hashes and length are optional. It also doesn't mention checking the length, if any, of the `snapshot` metadata.
* When we check the `targets` role or its delegations, do we mention checking optional hashes and length from the `snapshot` metadata?

lukpueh: Thanks for pointing these out, @trishankatdatadog!

> * 3.1 doesn't mention that hashes and length are optional. It also doesn't mention checking the length, if any, of the `snapshot` metadata.

Is another length match check in 3.1. necessary? We already say in 3. that the client should download snapshot.json up to the length specified in timestamp.json (if specified).

Regarding optionality of hashes, I agree we could add a "(if any)" note in 3.1, as we do for targets in 4.1.

> * When we check the `targets` role or its delegations, do we mention checking optional hashes and length from the `snapshot` metadata?

For top-level targets we do say in 4.1.   *"The hashes (if any), and  version number of the new targets metadata file MUST match the trusted snapshot metadata."* 
I think the "*(if any)*" is slightly misplaced, because the hashes it refers are computed by the client. It would be more accurate to say: *"The hashes and version number of the new targets metadata file MUST match the hashes (if any) and version number listed in the trusted snapshot"*. Let me push a commit that clarifies this.

Regarding length, similar to timestamp-snapshot in 3., we say in 4.1. that the client should download targets.json up to the length specified in snapshot.json (if specified). Do you think we need another length match check in 4.1.?

Regarding delegated targets metadata, we currently don't provide details about verification. #86 adds those details and adopts the [*"(if any)"* wording](https://github.com/theupdateframework/specification/pull/86/files#diff-94b2d37a08c587082c1ea1aaead6f973R1293-R1294) used in 4.1. If we accept my suggestion for 4.1. above, we should adopt this for delegated targets metadata in  #86.

Similarly, if we add a length match check for top-level targets metadata, we should adopt this for delegated in #86 too.

lukpueh: Clarified optionality of hashes (checks) in 0bdb99c. Let me know if we should add another optional length check...

mnm678: I think a second length check isn't necessary as the length is already used when downloading.

lukpueh:
Thanks for the spadework, @joshuagl, and for the thorough reviews, @mnm678, @trishankatdatadog and @JustinCappos. 

After having bumped last modified date and patch version number, as advised by our new [release policy](https://github.com/theupdateframework/specification#versioning), the checker script passes and we can merge and release v1.0.1. 🎉

tuf-spec.md
lukpueh: > When we check the targets role or its delegations, do we mention checking optional hashes and length from the snapshot metadata?
FYI, @trishankatdatadog: As promised above, I just pushed a commit to #86 that notes that delegated targets metadata hashes might be missing in snapshot metadata.

Issue #88, opened 2020-02-05
Reference the map file in the specification


vladimir-v-diaz:
[TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md) (the map file) describes how users may specify that a certain repository should be used for some targets, while other repositories are to be used for other targets.

We should reference the map file in the specification once TAP 4 is reviewed and accepted, which depends on the acceptance of TAP 7 (conformance testing).  Pending pull request for the map file implementation is available here: https://github.com/theupdateframework/tuf/pull/430

A detailed outline of the map file should be covered in the specification, similar to pull request theupdateframework/tuf#440 that has now been merged.

lukpueh:
As discussed in the TUF community meeting on  Dec 10, 2019 the changes of this PR require a minor bump of the spec version (see [meeting notes about "Versioning" ](https://groups.google.com/forum/#!msg/theupdateframework/oPTliXibT_4/0tJ_POLGAwAJ)).

joshuagl:
I'm planning to take a stab at this over the next week or two.

joshuagl:
For the interested, I submitted theupdateframework/taps#117 to try and clean up TAP 4 before I work on this.

joshuagl:
I haven't forgotten this work, but I want to discuss and resolve https://github.com/theupdateframework/specification/issues/109 on documenting concepts first.

PR #93, opened 2020-03-03
Remove TAP 3 + 4 adherence claim

lukpueh:
The specification claims that it adheres to TAP 3 and TAP 4, which is not the case. TAP 3 requires changes to the existing metadata format (prepared in #57) and TAP 4 adds a new metadata file (map.json), which is not yet mentioned in the specification (issue described in #88).

As agreed in the TUF community meeting on Dec 10, 2019, we remove TAP 3 and TAP 4 (adherence claims) for a formal 1.0.0 release and will add TAP 4 in version 1.1.x and TAP 3 in version 2.x.x.
(see [meeting notes about "Versioning" ](https://groups.google.com/forum/#!msg/theupdateframework/oPTliXibT_4/0tJ_POLGAwAJ))

*NOTE: Let's ignore the version management strategy and checker results for this PR. As agreed in the meeting (see notes above) this will be the last change before we formally release TUF 1.0.0. After that, we should stick to the rules defined in #87.*

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,6 +1,6 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **13 September 2019**
+Last modified: **5 March 2020**
 
 Version: **1.0.0**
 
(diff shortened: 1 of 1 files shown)


joshuagl:
I assume we will change the "Last modified" date before formally publishing spec version 1.0.0?

PR #94, opened 2020-03-03
Only suggest a release in check_release.py fo merges from draft

joshuagl:
A minor tweak to `check_release.py` to only suggest that the submitter tag a release and synch the draft branch with master when merging changes from the draft branch.


lukpueh:
Manually reverted this premature merge by hard resetting on master. We do, in the future, want to suggest release tags on every merge into master (if it bumps the version).

PR #98, opened 2020-05-04
Update snapshot role definition

sechkova:
**Issue:** Inconsistency in snapshot role definition throughout the specification

Snapshot role definition in 2.1.3 and 3.1.2 states that:

> The snapshot role signs a metadata file that provides information about the latest version of all of the other metadata on the repository  (excluding the timestamp file, discussed below).

While the detailed file formats description in 4.4 says: 

>  It MUST list the version numbers of the top-level targets metadata and all delegated targets metadata. 

From the latter, we understand that both `root.json` and `timestamp.json` are excluded from the snapshot which is also shown in the example files.

**Proposed changes:** 
Update snapshot role definition in 2.1.3 and 3.1.2 to match the  file format description (4.4) and explicitly state which metadata files must be listed in snapshot.json

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -331,8 +331,8 @@ repo](https://github.com/theupdateframework/specification/issues).
   - **2.1.3 Snapshot role**
 
       The snapshot role signs a metadata file that provides information about
-      the latest version of all of the other metadata on the repository
-      (excluding the timestamp file, discussed below).  This information allows
+      the latest version of all targets metadata on the repository
+      (the top-level targets.json and all delegated roles).  This information allows
       clients to know which metadata files have been updated and also prevents
       mix-and-match attacks.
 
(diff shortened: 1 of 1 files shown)


joshuagl:
FYI the Travis CI fails because the "Last modified:" and "Version:" fields in the header need to be updated before a change to the spec is merged. See the "[Versioning](https://github.com/theupdateframework/specification#versioning)" section of README.rst

sechkova:
Thanks for the review! 
"Version" and "Last modified" are updated :)

lukpueh:
> FYI the Travis CI fails because the "Last modified:" and "Version:" fields in the header need to be updated before a change to the spec is merged. See the "Versioning" section of README.rst

Hooray for the new release checker script being effective! :)

Issue #101, opened 2020-06-02
Remove JSON Requirement from the spec


mnm678:
In order to address concerns with canonical JSON (including #92) and to allow for more flexibility in the specification to make features like [TAP 11](https://github.com/theupdateframework/taps/blob/master/tap11.md) more useful, I propose we remove the canonical JSON requirement from the standard. This means removing the section:

> All documents use a subset of the JSON object format, with floating-point numbers omitted. When calculating the digest of an object, we use the "canonical JSON" subdialect as described at http://wiki.laptop.org/go/Canonical_JSON

In addition, we will likely have to do a pass over the document for consistency.

Note that this change does not affect any existing implementations. Canonical JSON will still be a valid choice, but future implementations may chose to use a different canonical json dialect or an entirely different metadata format while still following the TUF specification.

We can probably leave the examples in the spec as is, but make it clear that implementers may chose to use a different metadata format and still have a valid TUF implementation.

joshuagl:
Thanks for filing this @mnm678. We could adopt a similar approach to in-toto, whose spec makes it clear that there's no recommended format for the documents but that Canonical JSON is used in the specification to provide readable examples.

> To provide descriptive examples, we will adopt "canonical JSON," as described in http://wiki.laptop.org/go/Canonical_JSON, as the data format. However, applications that desire to implement in-toto are not required to use JSON. Discussion about the intended data format for in-toto can be found in the in-toto website.

joshuagl:
Note that the TUF website will need to be updated to reflect this change too, i.e. https://theupdateframework.io/metadata/
> All TUF metadata uses a subset of the JSON object format. When calculating the digest of an object, we use the Canonical JSON format.

mnm678:
Resolved in #102 and theupdateframework/theupdateframework.io#8

PR #102, opened 2020-06-03
Replace strict JSON requirement with a flexible requirement

mnm678:
Addresses #101 by replacing the JSON requirement with a statement that any data format may be used.

This pr keeps the canonical JSON examples in the specification, but adds a clarification about filenames to support non-json metadata.

This change is backwards compatible, so I propose it as a patch to the current version of the specification.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -412,7 +412,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
    The filenames and directory structure of repository metadata are strictly
    defined.  All metadata filenames will have an extension (EXT) based on the
-   metaformat, for example json metadata files would have an EXT of json.
+   metaformat, for example JSON metadata files would have an EXT of json.
    The following are the metadata files of top-level roles relative
    to the base URL of metadata available from a given repository mirror.
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -332,7 +332,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
       The snapshot role signs a metadata file that provides information about
       the latest version of all targets metadata on the repository
-      (the top-level targets.json and all delegated roles).  This information allows
+      (the top-level targets role and all delegated roles).  This information allows
       clients to know which metadata files have been updated and also prevents
       mix-and-match attacks.
 
@@ -498,8 +498,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
           KEYID is the identifier of the key signing the ROLE dictionary.
 
-          SIGNATURE is a hex-encoded signature of the canonical JSON form of
-          ROLE.
+          SIGNATURE is a hex-encoded signature of the metadata for ROLE.
 
 
    All keys have the format:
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
Great PR, Marina, thank you!

In fact, I think we should go further and remove strict tie-in with files and filesystems altogether...

mnm678:
> In fact, I think we should go further and remove strict tie-in with files and filesystems altogether...
I agree, but there are a lot of references to filesystems and filepaths throughout the spec, so it might be better to do that is a separate pr.

mnm678:
> Also, the whole spec is pretty json-heavy. So I wouldn't say that only the _examples_ here use json but also all of the _definitions_. Can we simply say that TUF doesn't have to be json, when all the metadata definitions are in json? I'm not sure if the json format can be unambiguously translated into any other format.

I worry that if we say the definitions are in json people will understand that as a json requirement. A translation may not be unambiguous, but I think it should be possible to include all the required in a different data format.

I removed the mentions of json in metadata definitions, so hopefully that helps.

lukpueh:
> I worry that if we say the definitions are in json people will understand that as a json requirement. A translation may not be unambiguous, but I think it should be possible to include all the required in a different data format.

Okay. Maybe we can extend the first sentence in 4.1. to say something like *"... as long as all fields in this specification are included and TUF clients are able to interpret them without ambiguity."?


> I removed the mentions of json in metadata definitions, so hopefully that helps.

Thanks! I think we should point out that the digest needs to be calculated in a deterministic manner by using the canonical form of whatever metadata format used, e.g. canonical JSON.

tuf-spec.md
```diff
@@ -411,34 +411,36 @@ repo](https://github.com/theupdateframework/specification/issues).
  + **3.1.2 Metadata files**
 
    The filenames and directory structure of repository metadata are strictly
-   defined.  The following are the metadata files of top-level roles relative
+   defined.  All metadata filenames will have an extension based on the
```
joshuagl, line 414:
```suggestion
   defined.  All metadata filenames will have an extension (EXT) based on the
```

mnm678:
Thanks for the review @lukpueh @joshuagl. I updated 4.1 to include a brief discussion about canonicalization.

tuf-spec.md
```diff
@@ -568,7 +575,7 @@ repo](https://github.com/theupdateframework/specification/issues).
         PUBLIC is in PEM format and a string.
 
    The KEYID of a key is the hexdigest of the SHA-256 hash of the
-   canonical JSON form of the key.
+   metadata format encoding of the key.
```
lukpueh, line 578: I'd prefer ... 
```suggestion
   canonical form of the key.
```
... akin to what you have for SIGNATURE. But I'm fine with this too.

mnm678:
@lukpueh I was thinking a patch release as this doesn't add/change functionality. If the text looks good I can update the version, date, etc

tuf-spec.md
```diff
@@ -411,7 +411,7 @@ repo](https://github.com/theupdateframework/specification/issues).
  + **3.1.2 Metadata files**
 
    The filenames and directory structure of repository metadata are strictly
-   defined.  All metadata filenames will have an extension based on the
+   defined.  All metadata filenames will have an extension (EXT) based on the
    metaformat, for example json metadata files would have an EXT of json.
```
joshuagl, line 415: Nit: We use JSON below, which is the correct name for the format.
```suggestion
   metaformat, for example JSON metadata files would have an EXT of json.
```

mnm678:
@lukpueh I'm having trouble getting travis-ci to run on this pr. Any idea what could be happening?

lukpueh:
Not sure what the problem was. But it looks like GitHub somehow had missed the notification from Travis. Just re-ran the build. Now it passes. Merging ...

PR #106, opened 2020-08-26
Timestamp rollback check should also check for snapshot rollback

erickt:
This updates section 5.2.2 to also state that when updating the `timestamp` role, we should also check to make sure new `timestamp` role's `snapshot` version is greater than or equal to the trusted `timestamp`'s `snapshot` version.

This is an important check when initializing a TUF database. For example, say we had the following TUF metadata:

* `root`, which has signed all the below metadata
* `snapshot` version 1, which points at a target with a compromised package
* `snapshot` version 2, which points at a target with a fixed package
* `timestamp` version 2, which points at snapshot version 2

Lets say an attacker has stolen the `timestamp` keys, they could forge a new `timestamp` role which points at `snapshot` version 1. In the normal case where we have a fully initialized TUF database with a trusted `root`, `timestamp` version 2, and `snapshot` version 2, this malicious timestamp would get rejected because of the section 5.3.2 snapshot rollback check.

However, consider a partially initialized repository, where we only trust `root` and `timestamp` version 2, and for some reason we're not able to fetch the rest of the metadata. In this setup, we won't have a trusted `snapshot` available for the 5.3.2 rollback check. Thus, the rollback attack would succeed.

To fix this, I've adopted the language in section 5.3.3, where when updating the `snapshot` role, we do explicitly check to see if the listed `targets` or any delegated `targets` version is always greater than or equal to the trusted versions of those roles.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,6 +1,6 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **9 June 2020**
+Last modified: **26 August 2020**
 
 Version: **1.0.4**
 
@@ -1160,7 +1160,7 @@ as FILENAME.EXT.
 
   * **2.2**. **Check for a rollback attack.**
 
-    * **3.3.1**. Note that the trusted timestamp metadata file may be checked
+    * **2.2.1**. Note that the trusted timestamp metadata file may be checked
     for authenticity, but its expiration does not matter for the following
     purposes.
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1160,17 +1160,13 @@ as FILENAME.EXT.
 
   * **2.2**. **Check for a rollback attack.**
 
-    * **2.2.1**. Note that the trusted timestamp metadata file may be checked
-    for authenticity, but its expiration does not matter for the following
-    purposes.
-
-    * **2.2.2**. The version number of the trusted timestamp metadata file, if
+    * **2.2.1**. The version number of the trusted timestamp metadata file, if
     any, must be less than or equal to the version number of the new timestamp
     metadata file.  If the new timestamp metadata file is older than the
     trusted timestamp metadata file, discard it, abort the update cycle, and
     report the potential rollback attack.
 
-    * **2.2.3**. The version number of the snapshot metadata file in the
+    * **2.2.2**. The version number of the snapshot metadata file in the
     trusted timestamp metadata file, if any, MUST be less than or equal to its
     version number in the new timestamp metadata file.  If not, discard the new
     timestamp metadadata file, abort the update cycle, and report the failure.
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1158,11 +1158,22 @@ as FILENAME.EXT.
...
   abort the update cycle, and report the signature failure.
 
-  * **2.2**. **Check for a rollback attack.** The version number of the trusted
-  timestamp metadata file, if any, must be less than or equal to the version
-  number of the new timestamp metadata file.  If the new timestamp metadata
-  file is older than the trusted timestamp metadata file, discard it, abort the
-  update cycle, and report the potential rollback attack.
+  * **2.2**. **Check for a rollback attack.**
+
+    * **3.3.1**. Note that the trusted timestamp metadata file may be checked
```
trishankatdatadog, line 1163: Do you mean **2.2.1** instead?

erickt: I did, thanks!

tuf-spec.md
```diff
@@ -1158,11 +1158,22 @@ as FILENAME.EXT.
...
-  * **2.2**. **Check for a rollback attack.** The version number of the trusted
-  timestamp metadata file, if any, must be less than or equal to the version
-  number of the new timestamp metadata file.  If the new timestamp metadata
-  file is older than the trusted timestamp metadata file, discard it, abort the
-  update cycle, and report the potential rollback attack.
+  * **2.2**. **Check for a rollback attack.**
+
+    * **2.2.1**. Note that the trusted timestamp metadata file may be checked
+    for authenticity, but its expiration does not matter for the following
+    purposes.
```
lukpueh, line 1165: Mildly off-topic (and I know that you only took over the wording form 3.3.*):

Do we really need this paragraph?  I find it a bit confusing/misleading. Authenticity not only *may* but *must* be and actually *is* checked in 2.1. Moreover, I'd rather say that the order of rollback check and expiration check doesn't matter, than saying that expiration does not matter. And, if at all, I wouldn't list this as a separate 2.2.1 workflow step, but rather as note on 2.2. The same concern applies to 3.3.* as well. What do others think?

lukpueh (commented):
Thanks for the addition and the detailed commit message, @erickt! This definitely makes sense. It would be nice if we could land it together with the strongly related #86, which clarifies/fleshes out rollback prevention for delegated targets in a similar fashion, and updates the corresponding parts about ffwd-recovery. (mind taking a look?). 

I left one only mildly related inline comment here, which is not a blocker for this PR. Still curious what others think.

tuf-spec.md
joshuagl: I'd agree with this, the paragraph doesn't appear to be necessary. If an updater is following the detailed workflow they have already checked the authenticity of the downloaded metadata. They will check the expiration of the newly downloaded metadata in the next step when checking for a freeze attack.

We might add it as a note, but I don't think even that is required. We don't usually talk about things we _aren't_ doing in the detailed workflow.

erickt: I'd be happy to remove it, I figured it was added to the snapshot rollback check for some reason I didn't understand :)

lukpueh: Excellent! Let's remove 2.2.1 then. I can send a follow up PR to remove 3.3.1 too. Or, if you like, you can do that as part of this PR, @erickt!

erickt: Thanks lukpueh! I updated the PR to remove 2.2.1 and 3.3.1.

lukpueh (approved):
Many thanks, @erickt! :)

Issue #92, opened 2020-02-26
Canonical JSON may not be valid JSON


daurnimator:
> All documents use a subset of the JSON object format, with floating-point numbers omitted. When calculating the digest of an object, we use the "canonical JSON" subdialect as described at http://wiki.laptop.org/go/Canonical_JSON

TIL that canonical json may invalid JSON: canonical json says that control characters *must not* be escaped:
 > Because only two byte values are escaped, be aware that JSON-encoded data may contain embedded control characters and nulls.

Whereas JSON mandates that control characters *are* escaped.

I think this deserves a note in the specification, as normal json encoders and decoders cannot be used.

joshuagl:
There are several related issues and discussions on Canonical JSON, i.e.
secure-systems-lab/securesystemslib#159 and theupdateframework/tuf#457

erickt:
I recently learned of another attempt at a canonicalization scheme in the draft IETF spec [rundgren-json-canonicalization-scheme](https://datatracker.ietf.org/doc/draft-rundgren-json-canonicalization-scheme/). It doesn't have the same issue where strings may contain invalid JSON characters. It's probably a drop-in replacement for most TUF implementations, which I think don't actually allow for invalid JSON characters.

lukpueh:
FYI: This topic has been added to the agenda for the next TUF community meeting. (date TBD, invitations will go to theupdateframework@googlegroups.com).

trishankatdatadog:
Yes. Please also join us on the CNCF Slack

PR #112, opened 2020-09-16
Checking version after verifying signatures

erickt:
In section 5.3.1 and section 5.4.1 of the spec, the first part of verifying a new snapshot, targets, and delegated targets role requires checking the new role's version number before we've checked that these new roles were signed by the root role. This exposes TUF clients to potential parser bugs which allows an attacker to potentially compromise the system.

For example, consider a man-in-the-middle attacker that has a parser bug which allows for executing arbitrary code, such as [CVE-2017-18349]. An attacker could exploit this in a few manners:

1. [POUF-1] inlines the metadata signatures in with the role JSON metadata file. While snapshot, targets, and delegated targets can have their hashes listed in trusted metadata (and thus be validated before parsing), the timestamp role must be parsed in order to verify it was signed by the trusted root role.
2. TUF-1.0.5 Section 5.3.1 of the workflow states the snapshot should be verified by hash and version number before checking the signature.  While we could verify the hash before parsing, the spec doesn't state that the hash should be checked first. It's possible then for a conformant TUF client to check the version number before hash, which would expose the client to the parser bug.
3. TUF-1.0.5 Section 4.4 and 5.4.1 make it optional for the snapshot role to contain the hashes of the targets and any delegated targets role. If left out, we need to still parse these files to check the version number before validating the signatures.

This patch addresses (2) and (3) by moving the verification of the version number to after we've verified a role was properly signed by the trusted root role. This would enable a future POUF that's addressed (1) to avoid future parser exploits.

[POUF-1]: https://github.com/theupdateframework/taps/blob/master/POUFs/reference-POUF/pouf1.md
[CVE-2017-18349]: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-18349

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -2,7 +2,7 @@
 
 Last modified: **30 September 2020**
 
-Version: **1.0.9**
+Version: **1.0.10**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
@@ -1191,11 +1191,12 @@ of the form VERSION_NUMBER.FILENAME.EXT (e.g., 42.snapshot.json), where
 VERSION_NUMBER is the version number of the snapshot metadata file listed in
 the timestamp metadata file.
 
-  * **5.3.1**. **Check against timestamp metadata.** The hashes and version
-  number of the new snapshot metadata file MUST match the hashes, if any, and
-  version number listed in the trusted timestamp metadata.  If hashes and
-  version do not match, discard the new snapshot metadata, abort the update
-  cycle, and report the failure.
+  * **5.3.1**. **Check against timestamp role's snapshot hash.** The hashes
+  of the new snapshot metadata file MUST match the hashes, if any, listed in
+  the trusted timestamp metadata.  This is done, in part, to prevent a
+  mix-and-match attack by man-in-the-middle attackers.  If the hashes do not
+  match, discard the new snapshot metadata, abort the update cycle, and report
+  the failure.
 
   * **5.3.2**. **Check for an arbitrary software attack.** The new snapshot
   metadata file MUST have been signed by a threshold of keys specified in the
@@ -1203,7 +1204,12 @@ the timestamp metadata file.
   as required, discard it, abort the update cycle, and report the signature
   failure.
 
-  * **5.3.3**. **Check for a rollback attack.** The version number of the targets
+  * **5.3.3**. **Check against timestamp role's snapshot version.** The version
+  number of the new snapshot metadata file MUST match the version number listed
+  in the trusted timestamp metadata.  If the versions do not match, discard the
+  new snapshot metadata, abort the update cycle, and report the failure.
+
+  * **5.3.4**. **Check for a rollback attack.** The version number of the targets
   metadata file, and all delegated targets metadata files, if any, in the
   trusted snapshot metadata file, if any, MUST be less than or equal to its
   version number in the new snapshot metadata file. Furthermore, any targets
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1188,35 +1188,40 @@ VERSION_NUMBER is the version number of the snapshot metadata file listed in
 the timestamp metadata file.  In either case, the client MUST write the file to
 non-volatile storage as FILENAME.EXT.
 
-  * **3.1**. **Check against timestamp metadata.** The hashes and version
-  number of the new snapshot metadata file MUST match the hashes (if any) and
-  version number listed in the trusted timestamp metadata.  If hashes and
-  version do not match, discard the new snapshot metadata, abort the update
-  cycle, and report the failure.
+  * **3.1**. **Check against timestamp role's snapshot hash.** The hashes
+  number of the new snapshot metadata file MUST match the hashes (if any)
```
joshuagl, line 1192:
```suggestion
 of the new snapshot metadata file MUST match the hashes (if any)
```

joshuagl (commented):
Making the spec more explicit about these steps to mitigate the impact of parser bugs is a solid change. Thanks @erickt!

I'm curious to understand whether you are doing something differently to POUF-1 in rust-tuf/go-tuf that makes (1) less of a problem for them?

erickt:
Thanks @joshuagl! As of right now, rust-tuf can't do much to avoid the parsing issues with POUF-1. We're relying on Rust and 
https://docs.serde.rs/serde_json/ to protect us from undefined behavior. We're mainly just leveraging changes like this to clean up rust-tuf in order to have better controls around exactly when deserialization happens. Then it'll become easier if/when we create a new POUF that addresses (1) (and other things, like the whole "canonical JSON isn't valid JSON" in https://github.com/theupdateframework/specification/issues/92).

joshuagl (approved):
This looks good to me.

mnm678 (approved):
LGTM. It looks like the reference implementation already checks the signature before the version in [`_update_metadata`](https://github.com/theupdateframework/tuf/blob/develop/tuf/client/updater.py#L1747)

lukpueh:
I cut a bunch of patch releases today that conflict with this PR. I have started [to rebase this PR on master, resolving conflicts (and squashing commits)](https://github.com/lukpueh/specification/compare/erickt/version-after-sig-orig..version-after-sig).

But the pending #119, which is ready to be merged, will create further conflicts. I suggest to first land #119 and then this PR. I'm happy to help resolving the new conflicts.

tuf-spec.md
```diff
@@ -1188,35 +1188,40 @@ VERSION_NUMBER is the version number of the snapshot metadata file listed in
 the timestamp metadata file.  In either case, the client MUST write the file to
 non-volatile storage as FILENAME.EXT.
 
-  * **3.1**. **Check against timestamp metadata.** The hashes and version
-  number of the new snapshot metadata file MUST match the hashes (if any) and
```
lukpueh, line 1192: I think we should keep the "*if any*", because snapshot's hashes in timestamp are still optional (see https://github.com/theupdateframework/specification/pull/90).

erickt: Ah I didn't realize the snapshot hash in timestamp was made optional. I'll restore this language after #119 lands.

tuf-spec.md
```diff
@@ -1233,48 +1238,53 @@ VERSION_NUMBER is the version number of the targets metadata file listed in the
...
+  not match, discard the new target metadata, abort the update cycle, and
+  report the failure.
 
   * **4.2**. **Check for an arbitrary software attack.** The new targets
   metadata file MUST have been signed by a threshold of keys specified in the
   trusted root metadata file.  If the new targets metadata file is not signed
   as required, discard it, abort the update cycle, and report the failure.
 
-  * **4.3**. **Check for a freeze attack.** The latest known time should be
+  * **4.3**. **Check against snapshot role's targets version.** The version
```
trishankatdatadog, line 1253: This is to check for a mix-and-match or rollback attack.

trishankatdatadog (commented):
LGTM provided we addressed all the comments

tuf-spec.md
lukpueh: Just rebased on top of master (post-#119) and restored the "if any" in 5.3.1.

tuf-spec.md
lukpueh: Do you want to add this here again?

lukpueh:
Just rebased on master, fixing conflicts and restoring the optionality of snapshot hashes in timestamp. The [diff to @erickt's original patch, including all the meanwhile merged PRs, can be found here](https://github.com/lukpueh/specification/compare/erickt/version-after-sig-orig..erickt:version-after-sig).

From my perspective this is good to go, but please double-check!

tuf-spec.md
trishankatdatadog: Yeah, I think just to say why we're doing this, because we try to do it for other checks.

joshuagl:
> From my perspective this is good to go, but please double-check!

Looks good to me.

PR #111, opened 2020-09-08
Remove slow retrieval attacks from protections

joshuagl:
I think we should remove slow retrieval attacks from 1.5.2. Goals to protect against specific attacks. I've included details in the commit message but the tl;dr is that:
1. AFAICT nothing in the _specification_ protects against these attacks
2. implementing protections against slow retrieval attacks has a high complexity for implementers which is not met by sufficient need
3. ownership of the network/download stack, in order to protect against slow retrieval attacks, is not something that clients are willing to give up (i.e. pip has thoroughly tested download code that they'd like to keep using)

PR #116, opened 2020-09-18
Persist metadata to local store after validation

erickt:
TUF-1.0.5 requires clients to write metadata to disk before it was validated, but this has multiple downsides:

* The metadata may have corrupted during the download. By writing it before validation, we prevent the client from using the local store.
* This could allow an attacker to persist malicious metadata that exploit a parser bug across reboots.

Instead, this patch changes the spec to explicitly write the metadata after validation to avoid these issues.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -2,7 +2,7 @@
 
 Last modified: **29 September 2020**
 
-Version: **1.0.7**
+Version: **1.0.8**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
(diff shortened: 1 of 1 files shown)


joshuagl (approved):
Yes, definitely. We recently had a discussion in the reference implementation about not persisting files before they have been verified (and have a related security advisory https://github.com/theupdateframework/tuf/security/advisories/GHSA-f8mr-jv2c-v8mg).

Thanks for codifying this in the spec!

lukpueh (approved):
LGTM! I took the liberty to rebase on the just updated (#117, #111 and d4c2f4b3d3707e1a2bfad22a02f1b5eb86dd6d18) master and update date/version header. Otherwise there were no conflicts, so I'm confident to go ahead and merge. Thanks for the contribution.

trishankatdatadog:
Sorry, just saw this. Wish we had clarified exactly when to persist _delegated_ targets metadata, although it may be obvious. We can always add this later!

trishankatdatadog:
Also, step 4.5 in Section 5 is already inconsistent, because it refers to the now outdated Sections `4.4.1 - 4.4.2.3`...

lukpueh:
> Sorry, just saw this. Wish we had clarified exactly when to persist delegated targets metadata, although it may be obvious. We can always add this later!

Feel free to PR! :P

> Also, step 4.5 in Section 5 is already inconsistent, because it refers to the now outdated Sections 4.4.1 - 4.4.2.3...

Yes, I realized this after merging. And fixed it as part of #119.

PR #117, opened 2020-09-23
Remove 2nd snapshot rollback attack check

erickt:
This shares the same justification for removal as #65. Step 3.3.1 was made redundant by #106, which modified the workflow to add 2.2.2, where updating the timestamp will also check if the new timestamp contains a snapshot version that is less than the trusted snapshot version.

This, in combination with the 3.1 check that asserts hashes and version of the actual snapshot metadata match the ones in the new trusted timestamp, make another version check, i.e, the one removed in this commit, obsolete.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **26 August 2020**
+Last modified: **23 September 2020**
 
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

lukpueh (approved):
Thanks for the patch and the detailed commit message, @erickt!

Issue #113, opened 2020-09-16
Should we check for timestamp role expiration when validating a new snapshot?


erickt:
It's possible there could be substantial amount of time between when updating the timestamp metadata, and updating a snapshot. So it's possible that at 5.2.3 the new timestamp metadata is unexpired, but it since expired by the time we start validating the new snapshot at 5.3. Does this matter? I ask because this is what rust-tuf does (see https://github.com/heartsucker/rust-tuf/blob/develop/src/tuf.rs#L223, look for all the calls to `safe_$ROLE_ref()`), and I'm trying to determine if this matters.

joshuagl:
A [discussion in Slack](https://cloud-native.slack.com/archives/C8NMD3QJ3/p1600294344057600) about this issue suggested that we could have a fixed notion of "_now_" for the duration of the update workflow, and that we should ensure that the delta between the current time and the fixed "_now_" should not be longer than the update period of timestamp.

rdimitrov:
@joshuagl - I think we can close this one considering #118 has been merged already 👍

joshuagl:
Yes, thanks Radoslav. This was resolved in #118.

PR #118, opened 2020-09-24
Introduce a fixed notion of time for the duration of the update cycle

joshuagl:
Addresses: #113 

Per the discussion between @erickt and @trishankatdatadog in Slack, this PR introduces a fixed notion of time to the client workflow. The client should store the time at the beginning of the update and use it for comparing expiration times throughout the detailed client workflow, so long as the session has not lasted longer than the re-signing period of the timestamp metadata.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1074,13 +1074,21 @@ repo](https://github.com/theupdateframework/specification/issues).
   still be able to update again in the future. Errors raised during the update
   process should not leave clients in an unrecoverable state.
 
-  **5.0**. **Load the trusted root metadata file.** We assume that a good,
+  **5.0**. **Record the time at which the update began.** Add the update
+  workflow maximum duration T to the recorded update start time to derive the
+  fixed update expiration time. The value for T is set by the authors of the
+  application using TUF. For example, T may be tens of minutes.
+  This update expiration time will be used when checking for freeze attacks,
+  and is fixed at the beginning of the update workflow to prevent metadata
+  from expiring during an in-progress update.
+
+  **5.1**. **Load the trusted root metadata file.** We assume that a good,
   trusted copy of this file was shipped with the package manager or software
   updater using an out-of-band process.  Note that the expiration of the
   trusted root metadata file does not matter, because we will attempt to update
   it in the next step.
 
-  **5.1**. **Update the root metadata file.** Since it may now be signed using
+  **5.2**. **Update the root metadata file.** Since it may now be signed using
   entirely different keys, the client MUST somehow be able to establish a
   trusted line of continuity to the latest set of keys (see Section 6.1). To do
   so, the client MUST download intermediate root metadata files, until the
@@ -1088,28 +1096,28 @@ repo](https://github.com/theupdateframework/specification/issues).
   consistent snapshots in order to download _versioned_ root metadata files as
   described next.
 
-  * **5.1.1**. Let N denote the version number of the trusted root metadata
+  * **5.2.1**. Let N denote the version number of the trusted root metadata
   file.
 
-  * **5.1.2**. **Try downloading version N+1 of the root metadata file**, up to
+  * **5.2.2**. **Try downloading version N+1 of the root metadata file**, up to
   some W number of bytes (because the size is unknown). The value for W is set
   by the authors of the application using TUF. For example, W may be tens of
   kilobytes. The filename used to download the root metadata file is of the
   fixed form VERSION_NUMBER.FILENAME.EXT (e.g., 42.root.json). If this file is
   not available, or we have downloaded more than Y number of root metadata
-  files (because the exact number is as yet unknown), then go to step 5.1.9.
+  files (because the exact number is as yet unknown), then go to step 5.2.9.
   The value for Y is set by the authors of the application using TUF. For
   example, Y may be 2^10.
 
-  * **5.1.3. Check for an arbitrary software attack.** Version N+1 of the root
+  * **5.2.3. Check for an arbitrary software attack.** Version N+1 of the root
   metadata file MUST have been signed by: (1) a threshold of keys specified in
   the trusted root metadata file (version N), and (2) a threshold of keys
   specified in the new root metadata file being validated (version N+1).  If
   version N+1 is not signed as required, discard it, abort the update cycle,
   and report the signature failure.  On the next update cycle, begin at step
-  5.0 and version N of the root metadata file.
+  5.1 and version N of the root metadata file.
 
-  * **5.1.4. Check for a rollback attack.** The version number of the trusted
+  * **5.2.4. Check for a rollback attack.** The version number of the trusted
   root metadata file (version N) MUST be less than or equal to the version
   number of the new root metadata file (version N+1). Effectively, this means
   checking that the version number signed in the new root metadata file is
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1119,21 +1125,29 @@ repo](https://github.com/theupdateframework/specification/issues).
...
 
-  * **1.7**. **Repeat steps 1.1 to 1.7**.
+  * **2.7**. **Repeat steps 1.1 to 1.7**.
+
+  * **2.8**. **Check for a freeze attack.**
+
+    * **2.8.1** The latest known time should be lower than the update start
+    time plus the update workflow maximum duration.  The update workflow
+    maximum duration is set by the authors of the application using TUF and
+    should typically be the same period at which the repository re-signs
```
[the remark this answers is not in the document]
JustinCappos: Also, I wonder if we need to keep repeating the suggestion about setting the update workflow maximum duration.  Might we be better off just saying this once?

tuf-spec.md
```diff
@@ -1076,41 +1077,46 @@ repo](https://github.com/theupdateframework/specification/issues).
   still be able to update again in the future. Errors raised during the update
   process should not leave clients in an unrecoverable state.
 
-  **0**. **Load the trusted root metadata file.** We assume that a good, trusted
+  **0**. **Record the time at which the update began.** This update start time
+  will be used when checking for freeze attacks.  Time is fixed at the
+  beginning of the update workflow to prevent metadata expiring during an
```
JustinCappos, line 1082:
```suggestion
  beginning of the update workflow to prevent metadata from expiring during an
```

JustinCappos (commented):
I want to discuss / think more about this.  I agree with the sentiment that we need to be more precise about how to handle time, but I want to make sure we're not opening up a can of worms in the slow retrieval case.

tuf-spec.md
```diff
@@ -29,7 +29,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 
    The keywords "MUST," "MUST NOT," "REQUIRED," "SHALL," "SHALL NOT," "SHOULD,"
    "SHOULD NOT," "RECOMMENDED," "MAY," and "OPTIONAL" in this document are to be
-   interpreted as described in RFC 2119.
+   interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
```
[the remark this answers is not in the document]
erickt: I'm just about to submit a typo-cleanup PR. I'll include these changes there.

tuf-spec.md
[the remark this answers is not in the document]
joshuagl: I wasn't entirely comfortable with this being the timestamp re-signing period. At first it seemed like a reasonable time period that was already defined by the system. However, it's not a period we define or discuss explicitly in the specification. 

Perhaps it makes more sense to have the update workflow maximum duration be explicitly application defined? I'd like it if we suggested a sane default value, but it might be more abstract. Like when we talk about downloading the root metadata and suggest

> up to some W number of bytes (because the size is unknown). The value for W is set by the authors of the application using TUF. For example, W may be tens of kilobytes.

maybe something like

> The latest known time should be lower than the update start time plus the update workflow maximum duration, D.  The value for D is set by the authors of the application using TUF. For example, D may be tens of minutes.

Agreed that it would be better to define the update workflow maximum duration in one place, we can do that at the same time as we record the beginning of the update in step 0. I will make that change.

lukpueh:
I cut a bunch of patch releases today that conflict with this PR. The pending #119 will create further conflicts. Give that I already cleaned up #119 so that it now merges cleanly with `master`, I suggest to first land #119 and then this PR. I am happy to then help resolving conflicts.

joshuagl:
Updated the PR to only introduce the fixed time changes, to remove redundant re-definition of the fixed update window and rebase on the recent changes.

tuf-spec.md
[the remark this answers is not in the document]
joshuagl: I've removed the explicit suggestion of using the signing period of the timestamp metadata and instead suggested that the time period is application defined.

joshuagl:
> I want to discuss / think more about this. I agree with the sentiment that we need to be more precise about how to handle time, but I want to make sure we're not opening up a can of worms in the slow retrieval case.

Hi @JustinCappos – did you have any further thoughts on this? Would you me to schedule a discussion to explore your concerns?

davidstrauss:
Coming from the side of PHP-TUF, we've gone with a library design that receives a timestamp for "now" when instantiated. This not only meets the goal of a consistent time for all sequential operations within the library but also allows better testing integration (e.g. with static fixtures) and framework integration (e.g. with ones like Drupal that already capture and retain a sense of "now" during each run that we can also use in TUF).

Any outer layer wrapping the library, like a CLI, unit test, or daemon, would then have responsibility for providing "now" to the library.

JustinCappos:
> > I want to discuss / think more about this. I agree with the sentiment that we need to be more precise about how to handle time, but I want to make sure we're not opening up a can of worms in the slow retrieval case.
> 
> Hi @JustinCappos – did you have any further thoughts on this? Would you me to schedule a discussion to explore your concerns?

I think I'm okay with this now.  In general, the concept of tracking the time at the beginning, having a window of time an update must complete in, etc. makes a lot of sense to me.

trishankatdatadog (approved):

trishankatdatadog:
Am I the only one who finds this language unintuitive on first read?

`The fixed update expiration time MUST be lower than the expiration timestamp in the current X metadata file.`

Would it be helpful to paraphrase it to something more intuitive? For example:

`The expiration timestamp in the current X metadata file MUST be higher than the current fixed notion of now.`

Cc @jhdalek55

joshuagl:
> Am I the only one who finds this language unintuitive on first read?

I suspect not. I tried to maintain the original phrasing of the freeze attack sections, but that was probably not necessary.

I couldn't think of a reason that "fixed now < expiration" was clearer than "expiration > fixed now", so I made the change you suggested. Thanks!

trishankatdatadog (dismissed):

joshuagl:
In the spirit of #132, could I get a second review/approval from @JustinCappos, @mnm678 or @lukpueh?

tuf-spec.md
```diff
@@ -1238,66 +1246,66 @@ of the form VERSION_NUMBER.FILENAME.EXT (e.g., 42.targets.json), where
...
+  abort the update cycle, and report the potential freeze attack.
 
-  * **5.4.5**. **Persist targets metadata.** The client MUST write the file to
+  * **5.5.5**. **Persist targets metadata.** The client MUST write the file to
   non-volatile storage as FILENAME.EXT (e.g. targets.json).
 
-  * **5.4.6**. **Perform a pre-order depth-first search for metadata about the
+  * **5.5.6**. **Perform a pre-order depth-first search for metadata about the
   desired target, beginning with the top-level targets role.**  Note: If
   any metadata requested in steps 5.4.6.1 - 5.4.6.2 cannot be downloaded nor
```
mnm678, line 1277:
```suggestion
  any metadata requested in steps 5.5.6.1 - 5.5.6.2 cannot be downloaded nor
```

mnm678 (dismissed):

tuf-spec.md
joshuagl: Well spotted, thank you! I've squashed that change into c2c4308

joshuagl:
I forced push to squash in some of the fixup commits introduced during code review, and that has dismissed the existing reviews.

Will need re-approval to meet the branch protection rules. Thank you in advance.

mnm678 (approved):

lukpueh:
@joshuagl, would you mind removing the merge commit form the PR?

joshuagl:
> @joshuagl, would you mind removing the merge commit form the PR?

Sure, done!

(I was trying to figure out whether I could avoid it as force pushes now dismiss reviews. Alas)

tuf-spec.md
```diff
@@ -1074,42 +1074,50 @@ repo](https://github.com/theupdateframework/specification/issues).
...
 
-  * **5.1.3. Check for an arbitrary software attack.** Version N+1 of the root
+  * **5.2.3. Check for an arbitrary software attack.** Version N+1 of the root
   metadata file MUST have been signed by: (1) a threshold of keys specified in
   the trusted root metadata file (version N), and (2) a threshold of keys
   specified in the new root metadata file being validated (version N+1).  If
   version N+1 is not signed as required, discard it, abort the update cycle,
   and report the signature failure.  On the next update cycle, begin at step
-  5.0 and version N of the root metadata file.
+  5.1 and version N of the root metadata file.
```
lukpueh, line 1118: There are two more "On the next update cycle, begin at step 5.0" below. Were they not changed for a reason?

tuf-spec.md
```diff
@@ -1239,66 +1247,66 @@ of the form VERSION_NUMBER.FILENAME.EXT (e.g., 42.targets.json), where
...
 
-      * **5.4.6.2.1**. If the current delegation is a multi-role delegation,
+      * **5.5.6.2.1**. If the current delegation is a multi-role delegation,
       recursively visit each role, and check that each has signed exactly the
       same non-custom metadata (i.e., length and hashes) about the target (or
       the lack of any such metadata).
 
-      * **5.4.6.2.2**. If the current delegation is a terminating delegation,
+      * **5.5.6.2.2**. If the current delegation is a terminating delegation,
       then jump to step 5.5.
```
lukpueh, line 1297:
```suggestion
      then jump to step 5.6.
```

lukpueh (changes requested):
The actual additions look good to me...

lukpueh:
I'm looking forward to #121, when we no longer have to manually keep track of section numbers.

tuf-spec.md
joshuagl: They were not. I thought I'd caught all of the cross-references. 🤦

lukpueh:
Btw. in [Section 7.2. we also reference "5.1"]( https://github.com/theupdateframework/specification/blame/890b156939116aaf4a0ffa7260a63be1a0c501a1/tuf-spec.md#L1437).

Judging from the git history, it means to [reference the section "The client application"](https://github.com/theupdateframework/specification/commit/daf0e4999abf3076fdcbed5648f09071d65b7cb8#diff-b0020aab85257910a5a31f248c201a81cb35e1939545c4b45515e9049e78cdffR979), which has no longer a numbering. Should we quickly fix that too?

joshuagl:
> reference the section "The client application"

Well spotted 👀 . I was going to make the reference link to "The client application" but decided against, because none of the other section references are hyperlinks and should will be easier to implement (and maintain) after #121

lukpueh (approved):
Thanks!! ❤️

lukpueh:
@mnm678, could you please re-approve?

mnm678 (approved):

trishankatdatadog:
For anyone interested, I tested [[1](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/0f51b6e10046e92726c893a5d79201e91ec897ec/src/tuf_on_a_plane/config.py#L40), [2](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/0f51b6e10046e92726c893a5d79201e91ec897ec/src/tuf_on_a_plane/repository.py#L51)] that this idea works 🙂

PR #120, opened 2020-09-25
Add TAP 11 to the specification.

mnm678:
Per the TAP workflow updated in theupdateframework/taps#124, this adds a description of POUFs to the specification to add support for TAP 11.

I added a short description of POUFs with a pointer to the TAP document for more details in order to keep the specification short. If others prefer, I could also add a longer description that includes the POUF document format and all of the other details from the TAP.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -2,7 +2,7 @@
 
 Last modified: **29 September 2020**
 
-Version: **1.0.8**
+Version: **1.1.0**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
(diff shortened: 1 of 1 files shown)


lukpueh:
Should this be a minor version change?

mnm678:
> Should this be a minor version change?

Yes, this 'adds functionality', so probably warrants a minor version change. I'll update the version number.

mnm678:
@lukpueh Travis is failing because the patch version was not incremented (but the minor version did increment). I think this is a Travis bug

lukpueh:
As per our release policy we don't merge minor version changes directly into master but into draft. Since this PR has master as destination the checker script thinks it's a patch level change.

mnm678:
I made a pr against draft in #122, and am closing here.

lukpueh:
FYI @mnm678, it's possible to change the base (destination) of a PR by clicking the edit button next to the PR title.

Issue #121, opened 2020-09-28
Rewriting the workflow to call out to sub-sections


erickt:
The TUF spec is written in a prose form, especially the update workflow. One of the challenges with this that we have duplicated paragraphs that describe essentially the same thing. For example:

> 2.1. Check signatures. The new timestamp metadata file must have been signed by a threshold of keys specified in the trusted root metadata file. If the new timestamp metadata file is not properly signed, discard it, abort the update cycle, and report the signature failure.
> ...
> 3.2. Check signatures. The new snapshot metadata file MUST have been signed by a threshold of keys specified in the trusted root metadata file. If the new snapshot metadata file is not signed as required, discard it, abort the update cycle, and report the signature failure.
> ...
> 4.2. Check for an arbitrary software attack. The new targets metadata file MUST have been signed by a threshold of keys specified in the trusted root metadata file. If the new targets metadata file is not signed as required, discard it, abort the update cycle, and report the failure.

Note each section has a slight change from each other, which happens as small corrections happen in one, but we forget to fix it in the others. Instead, I think it's worth considering restructuring the workflow (and possibly the rest of the doc) to be more in the style of the [whatwg](https://whatwg.org) specs (such as https://url.spec.whatwg.org/), where we call out to sub-sections to implement these major steps. For example, parsing the signatures could be written as:

> ## 2.1 Check the signatures.
> 1. Let _sig_result_ be the result of [check signatures](..) with _metadata_ being the new _timestamp_ metadata, _role_ being _timestamp_, and _root_ being the trusted root metadata.
> 2. If _sig_result_ is failure, then return failure.
> ...
> ## Check signatures:
> 
> The signature checker takes a metadata _metadata_, and a role type _role_, and a _root_ metadata.
> 
> 1. [Canonicalize the _metadata_](...) as _canonical_metadata_bytes_.
> 2. For each signature _sig_ in the _metadata_`'s _signature_ field:
>     1. If _sig_ is in a _processed_sig_ set, skip the signature.
>     2. If it signed _canonical_metadata_bytes_. Increment a _threshold_ counter.
> 3. If _threshold_ counter is less than _role_'s threshold in the _root_ metadata, return failure.

I think this sort of thing would really be helpful in making sure we don't make subtle copy-paste errors when changing the spec, and make it much easier on a client library to implement the spec correctly.

trishankatdatadog:
+1

lukpueh:
➕  from my side.
[...]

PR #122, opened 2020-09-29
Add TAP 11 to the specification

mnm678:
Add a description of POUFs, and a pointer to TAP 11 to add official support for TAP 11 to the specification.

supersedes #120.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **1 December 2020**
+Last modified: **2 December 2020**
 
-Version: **1.0.14**
+Version: **1.0.15**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
@@ -217,6 +217,8 @@ repo](https://github.com/theupdateframework/specification/issues).
             Mandatory Metadata signing schemes
         - [Tap 10](https://github.com/theupdateframework/taps/blob/master/tap10.md):
            Remove native support for compressed metadata
+        - [TAP 11](https://github.com/theupdateframework/taps/blob/master/tap11.md):
+           Using POUFs for Interoperability
 
 	   Implementations compliant with this version (1.0.0) of the specification
        must also comply with the TAPs mentioned above.
@@ -370,6 +372,20 @@ repo](https://github.com/theupdateframework/specification/issues).
    partly because the specific threat posted to clients in many situations is
    largely determined by how the framework is being used.
 
+*  **2.3. Protocol, Operations, Usage, and Format (POUF) Documents**
+
+    This specification purposefully leaves many implementation details,
+    including the metadata file formats, to the discretion of individual
+    implementations. These details do not affect the security of an
+    implementation, and so leaving them out of the specification allows this
+    document to support a greater variety of users. TUF implementers are
+    encouraged to document the wireline format and design decisions used in
+    their implementation as a POUF document. POUFs, as described in
+    [TAP 11](https://github.com/theupdateframework/taps/blob/master/tap11.md),
+    allow different adopters to create interoperable implementations of TUF.
+    POUFs should follow the layout described in TAP 11 and may be made
+    publicly available in the [TAP directory](https://github.com/theupdateframework/taps/tree/master/POUFs).
+
 ## **3. The repository**
 
    An application uses the framework to interact with one or more repositories.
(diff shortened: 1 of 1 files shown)


trishankatdatadog (approved):

tuf-spec.md
```diff
@@ -369,6 +371,20 @@ repo](https://github.com/theupdateframework/specification/issues).
    partly because the specific threat posted to clients in many situations is
    largely determined by how the framework is being used.
 
+*  **2.3. Protocol, Operations, Usage, and Format (POUF) Documents**
```
trishankatdatadog, line 374: On second thought, I think we should clarify that although we may concretely talk about files and use JSON as a pedagogical data format here, it is in no way an endorsement or a requirement.

tuf-spec.md
```diff
@@ -369,6 +371,20 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+
+    This specification purposefully leaves many implementation details,
+    including the metadata file formats, to the discretion of individual
+    implementations. These details do not affect the security of an
+    implementation, and so leaving them out of the specification allows this
+    document to support a greater variety of users. TUF implementers are
+    encouraged to document the wireline format and design decisions used in
+    their implementation is a POUF. POUFs, as described in
```
joshuagl, line 382:
```suggestion
    their implementation as a POUF document. POUFs, as described in
```

tuf-spec.md
joshuagl: Section 4.1 of the spec states:
> Implementers of TUF may use any data format for metadata files as long as all fields in this specification are included and TUF clients are able to interpret them without ambiguity. Implementers should choose a data format that allows for canonicalization, or one that will decode data deterministically by default so that signatures can be accurately verified. The examples in this document use a subset of the JSON object format, with floating-point numbers omitted. When calculating the digest of an object, we use the "canonical JSON" subdialect as described at http://wiki.laptop.org/go/Canonical_JSON

Do you think this needs further clarification?

joshuagl (approved):
LGTM, thanks @mnm678

joshuagl:
@mnm678 any reason this PR targets the draft branch and introduces a minor version bump? As a recommendation to document their protocols, operations, usage and formats; this doesn't seem to introduce any breaking changes for implementers.

lukpueh:
According to the Semantic Versioning Spec (https://semver.org/):
> MINOR version when you add functionality in a **backwards compatible** manner, and

lukpueh:
I'm unsure though if mentioning POUFs qualifies as "new functionality", IMO it rather is a refinement/clarification, which we might as well land on the patch level.

joshuagl:
Thanks for citing the SemVer spec @lukpueh, I should have written functional change. :-)

I don't _think_ we're adding any functionality to the spec with TAP 11/this PR. We are explaining that the spec intentionally leaves various details up to the implementer, and further recommending implementers create a POUF documenting their decisions. I think this would be fine to merge in master with a patch version increase?

mnm678:
@joshuagl In #120 we discussed POUFs as a new feature, but looking at it again, I agree with @lukpueh that they are more like additional documentation.

trishankatdatadog:
For an example of how a client may consume the same metadata but in different POUF data formats, the reader may wish to see this [demo](https://github.com/trishankatdatadog/tuf-on-a-plane/tree/0f51b6e10046e92726c893a5d79201e91ec897ec/src/tuf_on_a_plane/parsers)

lukpueh:
I just realized that it was me who suggested the minor bump in #120. Sorry for the confusion!  I suppose one could argue for both. But let's go with the refinement/clarification angle (see above) and merge it on the patch level. Changing the PR destination to `master` ...

joshuagl:
Excellent. I would love to land this before making any major changes to the spec structure (see #121).

Marina, could you rebase this PR and bump the version/date appropriately (perhaps after #135)?

mnm678:
I force pushed to rebase and squash the commits, so this should be good to go after #135

lukpueh (dismissed):
Just merged #135 and pushed another required conflict resolution here. IMO this now is indeed good to go.

lukpueh (approved):

lukpueh:
FYI, I also just migrated travis-ci.org to travis-ci.com...

joshuagl:
> FYI, I also just migrated travis-ci.org to travis-ci.com...

Thanks for doing that!

joshuagl (approved):
Looks good to me. 

@trishankatdatadog are you comfortable that the distinction of JSON being a pedagogical format, not a recommendation, is clear? See comment thread https://github.com/theupdateframework/specification/pull/122/files#r496919339

trishankatdatadog:
> Looks good to me.
> 
> @trishankatdatadog are you comfortable that the distinction of JSON being a pedagogical format, not a recommendation, is clear? See comment thread https://github.com/theupdateframework/specification/pull/122/files#r496919339

Yeah, I think it's fine, thanks, it should be clear enough from the document, but if not, maybe we should just add it to a FAQ somewhere.

PR #125, opened 2020-10-06
Minor change to clarify that 'custom' is an object

### LINKED ISSUE #50: Clarify the type of the target "custom" field
In 4.5 of the spec, the `TARGETPATH` is defined as:

```
 { TARGETPATH : {
       "length" : LENGTH,
       "hashes" : HASHES,
       ("custom" : { ... }) }
   , ...
 }
```

with custom being defined as:

> If defined, the elements and values of "custom" will be made available to the client application. The information in "custom" is opaque to the framework and can include version numbers, dependencies, requirements, and any other data that the application wants to include to describe the file at TARGETPATH. The application may use this information to guide download decisions.

Just to be safe, is `custom` a json object (as the snippet suggests), or is it allowed to be any json value? As best as I can tell, python-tuf asserts it's an [object](https://github.com/theupdateframework/tuf/blob/4f01a31346c2da9c0fb79a62cd6bb77da48ba5ed/tuf/formats.py#L231). If this the case, I'll submit a patch to make the spec more explicit.

joshuagl:
Minor tweak of the description of the custom field of TARGETPATH objects to explicitly state that the value is an object which is opaque to the framework.

Fixes #50

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **06 October 2020**
+Last modified: **07 October 2020**
 
-Version: **1.0.11**
+Version: **1.0.12**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
@@ -811,7 +811,7 @@ repo](https://github.com/theupdateframework/specification/issues).
        { TARGETPATH : {
              "length" : LENGTH,
              "hashes" : HASHES,
-             ("custom" : { ... }) }
+             ("custom" : CUSTOM) }
          , ...
        }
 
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -833,12 +833,12 @@ repo](https://github.com/theupdateframework/specification/issues).
...
    the hexdigest of the cryptographic function computed on the target file.
 
-   If defined, the elements and values of "custom" will be made available to the
-   client application.  The information in "custom" is opaque to the framework
-   and can include version numbers, dependencies, requirements, and any other
-   data that the application wants to include to describe the file at
-   TARGETPATH.  The application may use this information to guide download
-   decisions.
+   If defined, the elements and values of the CUSTOM object will be made
+   available to the client application.  The format of, and information in,
```
lukpueh, line 837: Minor nit: Do you think that it is clear that the CUSTOM object, if supplied, must be a json object (i.e. deserialize to a Python dictionary), if we say the "format ... is opaque"?

joshuagl: That's a good question. I wanted to try and avoid mentioning JSON in the description, because we are trying to make the actual format of the metadata looser in the specification.

I felt like the change here was as clear, from the prose alone, as other places we talk about objects. Though admittedly those other places in the specification where we talk about objects also have an explicit example of the object format and don't refer to it as opaque.

I will try and clarify this some more tomorrow. Either by clarifying the language here, by pointing out in section 4.1 that all objects in the file formats are JSON objects, or perhaps a combination.

lukpueh: I think it is not a very big deal. I guess my association with the word "format" here is biased through `securesystemslib.formats`, where the format also specifies the datatype. It's quite likely that others don't have that association.

joshuagl: I've pushed a fixup commit with some further clarification that the framework only needs to know that the value mapped to the "custom" attribute is an object.

lukpueh (approved):

joshuagl:
Force pushed an update to merge the fixup and increase the patch number to 11, assuming that this will land after #127

joshuagl:
In the spirit of #132, could I get a second review/approval @trishankatdatadog or @mnm678 ?

trishankatdatadog (approved):

PR #127, opened 2020-10-06
Minor clean-ups

joshuagl:
Fix a cross-reference and use consistent capitalisation for all titles.

I put the version and date bump in a separate patch, as it is not clear to me whether it is required for these very minor tweaks.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1309,7 +1309,7 @@ snapshot metadata file.
   (e.g.,
   c14aeb4ac9f4a8fc0d83d12482b9197452f6adf3eb710e3b1e2b79e8d14cb681.foobar.tar.gz),
   where HASH is one of the hashes of the targets file listed in the targets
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
Since there is no effective change, I don't even mind keeping it at the same spec version.

lukpueh:
I think we should either stick to our "no merge into master without release"-policy, or change the policy, but not make exceptions.
What would be the disadvantage of incrementing the patch version even for minor patches? (I think that's what the patch version number is for)

Alternatively we can merge into draft and aggregate it with other changes that will bump the minor version.

lukpueh (approved):
LGTM! Thanks :)

joshuagl:
In the spirit of #132, could I get a second review/approval @trishankatdatadog or @mnm678 ?

trishankatdatadog (approved):

PR #133, opened 2020-11-13
Add headers (and therefore anchor links) to points throughout the spec

xjm:
I frequently want to link someone to an exact point in the spec, especially in the client workfkow, and learned from discussion in the last meeting that the numbering of the points in the spec can change between versions. I thought a good fix to both issues would be to add sub-headers headers to the spec, which, in GitHub, automatically creates placeholder links that can be used to reference the exact step.

This PR only adds those headers, nested according to their depth under the top-level sections (which already had headers).

There were a few places in the client workflow where adding newlines and some slight additional indentation of the first line was necessary to retain correct formatting for the paragraphs describing each step.

For the cleanest source, these paragraphs should all be re-indented to the same depth as the first line of the paragraph and then rewrapped at what appears to be 80 chars, but I didn't want to include that change here yet because it makes a mess of the diff.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -23,7 +23,7 @@ repo](https://github.com/theupdateframework/specification/issues).
 - [F. Future Directions and Open Questions](#f-future-directions-and-open-questions)
 
 ## **1. Introduction**
-* **1.1. Scope**
+* ### **1.1. Scope**
 
    This document describes a framework for securing software update systems.
 
@@ -31,7 +31,7 @@ repo](https://github.com/theupdateframework/specification/issues).
    "SHOULD NOT," "RECOMMENDED," "MAY," and "OPTIONAL" in this document are to be
    interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
-* **1.2. Motivation**
+* ### **1.2. Motivation**
 
    Software is commonly updated through software update systems.  These systems
    can be package managers that are responsible for all of the software that is
@@ -49,7 +49,7 @@ repo](https://github.com/theupdateframework/specification/issues).
    Software update systems are vulnerable to a variety of known attacks.  This
    is generally true even for implementations that have tried to be secure.
 
-* **1.3. History and credit**
+* ### **1.3. History and credit**
 
    Work on TUF began in late 2009.  The core ideas are based off of previous
    work done by Justin Cappos and Justin Samuel that [identified security flaws
@@ -75,7 +75,7 @@ repo](https://github.com/theupdateframework/specification/issues).
    work of Nick Mathewson, Sebastian Hahn, Roger Dingledine, Martin Peck, and
    others.
 
- * **1.4. Non-goals**
+ * ### **1.4. Non-goals**
 
    We are not creating a universal update system, but rather a simple and
    flexible way that applications can have high levels of security with their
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1117,25 +1121,28 @@ repo](https://github.com/theupdateframework/specification/issues).
...
+  * #### **5.1.7**. **Persist root metadata.**
+      The client MUST write the file to
   non-volatile storage as FILENAME.EXT (e.g. root.json).
 
-  * **5.1.8**. **Repeat steps 5.1.1 to 5.1.8**.
+  * #### **5.1.8**. **Repeat steps 5.1.1 to 5.1.8**.
 
-  * **5.1.9**. **Check for a freeze attack.** The latest known time MUST be
+  * #### **5.1.9**. **Check for a freeze attack.**
+      The latest known time MUST be
```
trishankatdatadog, line 1138: The nonalignment with the header looks odd... is it necessary here and in other places where it happens, too?

xjm: @trishankatdatadog Yep. If I remove the early wrapping on this line and accordingly rewrap the paragraph at the new indentation, that results in an illegible diff wherein it appears I've rewritten every single word of the client spec, when it's only a whitespace change. ;) So I figured better to let folks review some slightly odd whitespace that doesn't actually have any visual impact, and then clean up the source wrapping later.

xjm:
Note that it's also worth looking at this live in my fork, to see the rendered whitespace and permalinks: https://github.com/xjm/specification/blob/main/tuf-spec.md

trishankatdatadog:
This is great, thanks! The anchors don't always follow [Fitt's Law](https://en.wikipedia.org/wiki/Fitts%27s_law), but that's not something we can really control, I think.

I have a suggestion: would you please fix sections like 5.1.11, where the text mistakenly breaks from the bold styling? If we don't have short section headers/titles as anchors, we should make them.

joshuagl:
We will get anchor links for free with the [proposed change](https://github.com/theupdateframework/specification/pull/143) to use Bikeshed, i.e. https://joshuagl.github.io/specification/#load-trusted-root

trishankatdatadog:
> We will get anchor links for free with the [proposed change](https://github.com/theupdateframework/specification/pull/143) to use Bikeshed, i.e. https://joshuagl.github.io/specification/#load-trusted-root

Looks great! Are we Bikeshedding anytime soon, Joshua? That will tell us whether we should merge this PR (once conflicts are fixed) for now or not.

joshuagl:
> > We will get anchor links for free with the [proposed change](https://github.com/theupdateframework/specification/pull/143) to use Bikeshed, i.e. https://joshuagl.github.io/specification/#load-trusted-root
> 
> Looks great! Are we Bikeshedding anytime soon, Joshua? That will tell us whether we should merge this PR (once conflicts are fixed) for now or not.

I have opened a PR with initial bikeshed conversion of the specification here https://github.com/theupdateframework/specification/pull/143

trishankatdatadog:
> I have opened a PR with initial bikeshed conversion of the specification here #143

@xjm is it OK if we close this PR _after_ we merge #143?

joshuagl:
Thank you for the PR @xjm. We have since landed a conversion of the specification to Bikeshed flavoured markdown (#143) and are generating a HTML specification which includes anchor links for the high-level steps of the [detailed client workflow](https://theupdateframework.github.io/specification/latest/#detailed-client-workflow).

The anchors in the bikeshed generated specification are not as granular as the anchors you created here, but we would like them to be. We may want to tackle this at the same time as #146 or even #121, though the amount of work on either of those issues may be more than you have the time or inclination to do. Please let us know how you would like to proceed with this PR.

joshuagl:
Closing this because it no longer applies after the bikeshed conversion and discussion here has gone quiet.

PR #135, opened 2020-11-30
Fix METAFILES definition to match the text

### LINKED ISSUE #134: METAFILES object format confusion
spec defines METAFILES object format as 
```
 { METAPATH : {
       "version" : VERSION,
       ("length" : LENGTH, |
        "hashes" : HASHES) }
   , ...
 }
```
The text makes it clear that both length and hashes are optional (and not dependent on each other) but the construct `(name:value, | name:value)` seems to imply that exactly one of these should be specified: this is how DELEGATIONS uses the same construct.

Object format should maybe be changed to
```
 { METAPATH : {
       "version" : VERSION,
       ("length" : LENGTH,)
       ("hashes" : HASHES) }
   , ...
 }
```
or is this is a misunderstanding of the definition format?

jku:
The text makes it clear that both "length" and "hashes" are optional and
not dependent on each other. The object format seems to imply that
both should not be included at the same time: DELEGATIONS uses the same
construct to express exactly that.

Express more clearly that the attributes are not dependent on each
other.

---
Fixes #134

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **25 November 2020**
+Last modified: **1 December 2020**
 
-Version: **1.0.13**
+Version: **1.0.14**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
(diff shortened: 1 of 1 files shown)


lukpueh:
Thanks, @jku! Would you mind bumping the last modified date and patch version number in the document head?

joshuagl (approved):
Thanks @jku!

mnm678 (approved):

PR #136, opened 2020-12-01
Fix freeze attack checks after the introduction of workflow maximum duration (by removing workflow maximum duration?)

joshuagl:
When introducing the fixed notion of time I also added "workflow maximum duration T" to allow for some slack in the expiration time, but the way T is combined with the fixed notion of time results in what should be a valid expiration being classed as expired and TUF stopping the update for a possible freeze attack. There are more details in the commit message.

Thanks to @trishankatdatadog for spotting this inconsistency.

The simplest way to clean this up was to remove the fixed time slack, T. Especially as the purpose of the PR was to introduce the fixed notion of time, the slack was feature creep.

However, I'd be interested in hearing from TAP editors on the following:
* Is the slack time (T) a useful notion to re-introduce (along with updated descriptions of the freeze attack checks)? @trishankatdatadog observed it could be useful to allow for things like NTP drift.
* Should we change the expiration comparison from 

> "expiration timestamp in the [...] metadata file MUST be higher than the fixed update expiration time"

to 

> "expiration timestamp in the [...] metadata file MUST be higher than, or equal to, the fixed update expiration time"

? (Either way, #121 will help massively here)

I plan to update this PR if we decide to include T and/or change how we describe expiration comparison for freeze attacks.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1093,9 +1093,9 @@ repo](https://github.com/theupdateframework/specification/issues).
 
   **5.0**. **Record the time at which the update began** as the fixed update
   start time.  Time is fixed at the beginning of the update workflow to allow
-  an application using TUF to effectively pause time, in order to prevent
-  metadata which is valid at the beginning of an update from expiring during
-  the update workflow.
+  an application using TUF to effectively pause time, in order to ensure that
+  metadata which has a valid expiration time at the beginning of an update
+  does not fail an expiration check later in the update workflow.
 
   **5.1**. **Load the trusted root metadata file.** We assume that a good,
   trusted copy of this file was shipped with the package manager or software
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **2 December 2020**
+Last modified: **3 December 2020**
 
-Version: **1.0.15**
+Version: **1.0.16**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
@@ -1091,13 +1091,11 @@ repo](https://github.com/theupdateframework/specification/issues).
   still be able to update again in the future. Errors raised during the update
   process should not leave clients in an unrecoverable state.
 
-  **5.0**. **Record the time at which the update began.** Add the update
-  workflow maximum duration T to the recorded update start time to derive the
-  fixed update expiration time. The value for T is set by the authors of the
-  application using TUF. For example, T may be tens of minutes.
-  This update expiration time will be used when checking for freeze attacks,
-  and is fixed at the beginning of the update workflow to prevent metadata
-  from expiring during an in-progress update.
+  **5.0**. **Record the time at which the update began** as the fixed update
+  expiration time.  Time is fixed at the beginning of the update workflow to
+  allow an application using TUF to effectively pause time, in order to prevent
+  metadata which is valid at the beginning of an update from expiring during
+  the update workflow.
 
   **5.1**. **Load the trusted root metadata file.** We assume that a good,
   trusted copy of this file was shipped with the package manager or software
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1092,8 +1092,8 @@ repo](https://github.com/theupdateframework/specification/issues).
   process should not leave clients in an unrecoverable state.
 
   **5.0**. **Record the time at which the update began** as the fixed update
-  expiration time.  Time is fixed at the beginning of the update workflow to
-  allow an application using TUF to effectively pause time, in order to prevent
+  start time.  Time is fixed at the beginning of the update workflow to allow
+  an application using TUF to effectively pause time, in order to prevent
   metadata which is valid at the beginning of an update from expiring during
   the update workflow.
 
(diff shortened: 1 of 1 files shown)


trishankatdatadog:
Thanks very much, @joshuagl 🙂 I agree that removing the slack time is probably the most prudent move for now.

Should we also discuss whether for all freeze attack checks, "the expiration timestamp in the new X metadata file MUST be ~higher than~ **greater than or equal to** the fixed update expiration time", or should that be another issue/PR?

trishankatdatadog:
So, just to clarify, I think the pseudocode should be:

```
fixed_time = now(UTC) - T
# language in the spec
not_expired = expiration > fixed_time
# same thing said different
expired = fixed_time < expiration
```

T is optional and zero by default. So the question now is whether we add equality to the time comparisons.

joshuagl:
> So, just to clarify, I think the pseudocode should be:
> 
> ```
> fixed_time = now(UTC) - T
> # language in the spec
> not_expired = expiration > fixed_time
> # complement
> expired = expiration <= fixed_time
> ```
> 
> T is optional and zero by default.

Thanks for the clarification with pseudocode, that matches my understanding too. With T being removed in this PR. 
I've filed https://github.com/theupdateframework/specification/issues/137 to discuss re-introducing T in future.

I've also updated this PR to bump version and last modified, so that there's a clean merge. I had to fib about the last modified date (set to tomorrow) to make the CI check pass...

trishankatdatadog:
> I've also updated this PR to bump version and last modified, so that there's a clean merge. I had to fib about the last modified date (set to tomorrow) to make the CI check pass...

This is so meta... 😂

lukpueh:
> I had to fib about the last modified date (set to tomorrow) to make the CI check pass...

That's odd, I only just recently fixed this in the release checker script: https://github.com/theupdateframework/specification/commit/d4c2f4b3d3707e1a2bfad22a02f1b5eb86dd6d18

joshuagl:
> > I had to fib about the last modified date (set to tomorrow) to make the CI check pass...
> 
> That's odd, I only just recently fixed this in the release checker script: https://github.com/theupdateframework/specification/commit/d4c2f4b3d3707e1a2bfad22a02f1b5eb86dd6d18

Oh, I fibbed twice. I'd remembered the old behaviour and completely missed your fix. Thanks @lukpueh.

I have force pushed today's date.

trishankatdatadog (commented):
LGTM but I'd like other people to independently review just to make sure 🙂 Thanks, Joshua!

tuf-spec.md
```diff
@@ -1091,13 +1091,11 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-  fixed update expiration time. The value for T is set by the authors of the
-  application using TUF. For example, T may be tens of minutes.
-  This update expiration time will be used when checking for freeze attacks,
-  and is fixed at the beginning of the update workflow to prevent metadata
-  from expiring during an in-progress update.
+  **5.0**. **Record the time at which the update began** as the fixed update
+  expiration time.  Time is fixed at the beginning of the update workflow to
+  allow an application using TUF to effectively pause time, in order to prevent
+  metadata which is valid at the beginning of an update from expiring during
+  the update worfklow.
```
mnm678, line 1098:
```suggestion
  the update workflow.
```

mnm678 (approved):
One small typo, but his looks good.

I think that some notion of slack time would be a good addition in the future. It may also be useful for offline clients that may need a wider expiration window.

joshuagl:
> One small typo, but his looks good.

Thanks Marina. I really need to use an editor with spell-checking enabled when making even small changes to the spec.

> I think that some notion of slack time would be a good addition in the future. It may also be useful for offline clients that may need a wider expiration window.

That's a good motivation also, I've added that to #137.

joshuagl:
> > One small typo, but his looks good.
> 
> Thanks Marina. I really need to use an editor with spell-checking enabled when making even small changes to the spec.

I force pushed a typo fix, which has dismissed your original review. Would appreciate a re-review when you have time @mnm678. Thanks!

trishankatdatadog:
Thanks, Joshua and Marina!

While we are at it, should we also fix https://github.com/theupdateframework/tuf/issues/1231 or push that to another PR, if need be?

joshuagl:
> While we are at it, should we also fix [theupdateframework/tuf#1231](https://github.com/theupdateframework/tuf/issues/1231) or push that to another PR, if need be?

Probably push to another PR while we try and build consensus?

tuf-spec.md
```diff
@@ -1091,13 +1091,11 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-  fixed update expiration time. The value for T is set by the authors of the
-  application using TUF. For example, T may be tens of minutes.
-  This update expiration time will be used when checking for freeze attacks,
-  and is fixed at the beginning of the update workflow to prevent metadata
-  from expiring during an in-progress update.
+  **5.0**. **Record the time at which the update began** as the fixed update
+  expiration time.  Time is fixed at the beginning of the update workflow to
+  allow an application using TUF to effectively pause time, in order to prevent
+  metadata which is valid at the beginning of an update from expiring during
+  the update workflow.
```
lukpueh, line 1098: Similarly, I'm unsure about the causality implied in "pause time, in orderer to prevent". Don't we want to say something like this...?

```suggestion
  allows an application using TUF to effectively pause time, without 
  metadata, which is valid at the beginning of an update, becoming expired during
  the update workflow.
```

lukpueh (changes requested):
Thanks for the patch, @joshuagl. I have two nit picks, to be taken with a non-native-grain of salt. Please disregard if unjustified.

tuf-spec.md
```diff
@@ -1091,13 +1091,11 @@ repo](https://github.com/theupdateframework/specification/issues).
...
-  **5.0**. **Record the time at which the update began.** Add the update
-  workflow maximum duration T to the recorded update start time to derive the
-  fixed update expiration time. The value for T is set by the authors of the
-  application using TUF. For example, T may be tens of minutes.
-  This update expiration time will be used when checking for freeze attacks,
-  and is fixed at the beginning of the update workflow to prevent metadata
-  from expiring during an in-progress update.
+  **5.0**. **Record the time at which the update began** as the fixed update
+  expiration time.  Time is fixed at the beginning of the update workflow to
+  allow an application using TUF to effectively pause time, in order to prevent
```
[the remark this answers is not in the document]
joshuagl: You are absolutely correct. I'll replace with fixed update start time, this term is from an earlier attempt at defining this process. Well caught, thank you.

tuf-spec.md
joshuagl: How about:

> pause time, in order to ensure that metadata which has a valid expiration time at the beginning of the update does not fail an expiration check later in the update workflow

?

lukpueh: I just realized that I mistook "pause time" for "pause execution", so the original phrasing does make sense and so does your update. I withdraw my suggestion and am fine with both of yours.

lukpueh (approved):
🎉  Thanks for addressing my concerns! LGTM.

mnm678 (approved):

PR #141, opened 2020-12-11
Remove multi-role vestigial

### LINKED ISSUE #140: Is multi-role delegation in 1.0.0 or not?
In the [current](https://github.com/theupdateframework/specification/blob/fd85a8a12403435d84e202ea8790eb4abad4559d/tuf-spec.md#5-detailed-workflows) version of the spec, we say:

> 5.5.6.2.1. If the current delegation is a multi-role delegation, recursively visit each role, and check that each has signed exactly the same non-custom metadata (i.e., length and hashes) about the target (or the lack of any such metadata).

But nowhere else do we actually define how to specify multi-role delegations. Did we: (1) forget to specify how multi-role delegations are listed (since they are [backwards-incompatible](https://github.com/theupdateframework/taps/blob/cbea9b402e597e59a37f7caf73383fb5adde409d/tap3.md#backwards-compatibility)), or (2) did we agree not to release them as part of 1.0.0, and this is leftover text intended for 2.0.0?

trishankatdatadog:
Fix #140 

Does this adhere to rules set in #87?

Signed-off-by: Trishank Karthik Kuppusamy <trishank.kuppusamy@datadoghq.com>

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1,8 +1,8 @@
 # <p align="center">The Update Framework Specification
 
-Last modified: **3 December 2020**
+Last modified: **11 December 2020**
 
-Version: **1.0.16**
+Version: **1.0.17**
 
 We strive to make the specification easy to implement, so if you come across
 any inconsistencies or experience any difficulty, do let us know by sending an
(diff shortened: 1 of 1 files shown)


joshuagl:
I rebased this PR on top of the bikeshed changes here: https://github.com/joshuagl/specification/tree/trishankatdatadog/remove-tap-3-vestigial

joshuagl:
I've pushed a new PR #151 which rebases this patch following the bikeshed conversion.

Issue #91, opened 2020-02-14
Secondary literature with detailed rationale and recommendations


joshuagl:
It could be valuable for potential adopters of TUF if there were some documentation beyond the specification, published papers and conversations captured on GitHub, that goes into detail about certain decisions, makes recommendations where the specification deliberately leaves things open and points to open implementations of the specification (i.e. Notary and PEP 458) as examples of the context for the various decisions that must be made when applying the TUF specification to a scenario.

The spec is a good document but provides several points where choices must be made without providing any explanation or guidance.

The papers which motivated various spec decisions and changes provide interesting reading but can be a little dense when trying to understand a nuance of the specification where the context for a decision may be difficult to elicit and, furthermore, the papers are a static document, unlike the specification itself.

In contrast to the specification, which should only say “do XYZ”, this additional document could say things like “do XYZ because foo, bar, baz” or “do X if your situation is quux (akin to projects ABC) or do Y if it is thud (akin to projects DEF)”.

cc @lukpueh

JustinCappos:
This is a good suggestion.  We have a document like this for the automotive
variant of TUF (Uptane) called the Deployment Considerations ( e.g., see
part of it here:
https://uptane.github.io/deployment-considerations/repositories.html ).

We should think about how to get relevant information back into TUF.

lukpueh:
Also see secondary literature worthy [discussion](https://github.com/theupdateframework/specification/pull/90#discussion_r387254029)  by @mnm678 and @trishankatdatadog about when to use (not use) optional length/hashes fields in timestamp and snapshot metadata.

joshuagl:
Trishank shared some wisdom during a PEP 458/Warehouse integration discussion that I think is worth capturing as a recommendation in the secondary literature.

Paraphrasing his advice: 
> keep pinned trusted copy of the root metadata, at least the hash, in the repository code so that there's a trusted base to build from

@trishankatdatadog please correct me if I didn't quite capture that right.

trishankatdatadog:
> @trishankatdatadog please correct me if I didn't quite capture that right.

Absolutely right, thank you 🙏

joshuagl:
Other interesting secondary literature topics:
1. how to initialise trust from a local copy of the metadata #108 
2. techniques for recovering from an interrupted update #69, or how to interpret the following statement from the detailed client workflow:
[...]

PR #142, opened 2020-12-11
Explain why we check hashes before signatures

### LINKED ISSUE #138: Explain why checking some hashes before signatures is ok?
From this [PR on php-tuf](https://github.com/php-tuf/php-tuf/pull/88#issuecomment-737045948), I learned that it is not clear why the spec advises on checking snapshot/targets hashes from timestamp/snapshot before checking signatures. It looks unsafe until it is recalled that the hashes from previously verified snapshot/targets. Recently while independently implementing the spec, even I had some concern until I understood this. Should we clarify this?

HT @tedbow @davidstrauss

trishankatdatadog:
An attempt to fix #138 

Does this look good @tedbow?

Signed-off-by: Trishank Karthik Kuppusamy <trishank.kuppusamy@datadoghq.com>

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1400,7 +1400,7 @@ it in the next step.
   of the new snapshot metadata file MUST match the hashes, if any, listed in
   the trusted timestamp metadata.  This is done, in part, to prevent a
   mix-and-match attack by man-in-the-middle attackers. It is safe to check the
-  hashes before the signatures, because the hashes comes from the timestamp
+  hashes before the signatures, because the hashes come from the timestamp
   role, which we have already verified in the previous step; it is also a quick
   way to reject bad metadata. If the hashes do not match, discard the
   new snapshot metadata, abort the update cycle, and report the failure.
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.24
+Text Macro: VERSION 1.0.25
 </pre>
 
 Note: We strive to make the specification easy to implement, so if you come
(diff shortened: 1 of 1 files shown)


lukpueh (commented):
Thanks for clarifying this, @trishankatdatadog. I wonder if we can make it more concise.

>It is safe to check the hashes before the signatures, because the hashes comes from the timestamp role, which we have already verified in the previous step

feels a bit redundant to 
>  hashes ... listed in the trusted timestamp metadata

Maybe it's enough to just expand what's already there, e.g.  
> hashes ... listed in the above/previously/already verified and thus trusted timestamp metadata

?


On the other hand, it seems worthwhile to elaborate on the 
> quick way to reject bad metadata

What about something along the lines of

> preliminary integrity check before performing a more expensive signature verification

?

Besides, should this be part of the spec or is it rather a contender for secondary literature (https://github.com/theupdateframework/specification/issues/91)?

trishankatdatadog:
Sounds good, I will address your comments. In the meantime, I think we should add this to the spec, because that's where the question came up for php-tuf while implementing, but I'll let @tedbow decide.

joshuagl:
I agree that the spec is a good place for this, at least until we have the secondary literature. There are other parts of the spec where we explain why things are they way they are, such as the recently introduced fixed update start time.

trishankatdatadog:
@joshuagl @mnm678 PTAL 👀

tuf-spec.md
```diff
@@ -1448,9 +1450,11 @@ it in the next step.
 2. **Check against snapshot role's targets hash**. The hashes
   of the new targets metadata file MUST match the hashes, if any, listed in the
   trusted snapshot metadata.  This is done, in part, to prevent a mix-and-match
-  attack by man-in-the-middle attackers.  If the new targets metadata file does
-  not match, discard the new target metadata, abort the update cycle, and
-  report the failure.
+  attack by man-in-the-middle attackers. It is safe to check the hashes before
+  the signatures, because the hashes comes from the snapshot role, which we have
```
joshuagl, line 1454:
```suggestion
  the signatures, because the hashes come from the snapshot role, which we have
```

tuf-spec.md
```diff
@@ -1399,9 +1399,11 @@ it in the next step.
 2. **Check against timestamp role's snapshot hash**. The hashes
   of the new snapshot metadata file MUST match the hashes, if any, listed in
   the trusted timestamp metadata.  This is done, in part, to prevent a
-  mix-and-match attack by man-in-the-middle attackers.  If the hashes do not
-  match, discard the new snapshot metadata, abort the update cycle, and report
-  the failure.
+  mix-and-match attack by man-in-the-middle attackers. It is safe to check the
+  hashes before the signatures, because the hashes comes from the timestamp
```
joshuagl, line 1403:
```suggestion
  hashes before the signatures, because the hashes come from the timestamp
```

joshuagl (approved):
LGTM, thanks Trishank

joshuagl (approved):

mnm678 (approved):

PR #150, opened 2021-02-25
Rollback attacks and fast forward recovery

mnm678:
This is my attempted takeover of #86. Based on discussion in that pr, I decided to explicitly recommend that fast forward attack recovery for delegated targets roles requires replacement of snapshot keys. Specifically this pr:
* Removes redundant rollback protection by leaving #65 in place and removes the redundant rollback check for delegated targets.
* Adds a note about fast forward attack recovery for delegated targets.
* Includes related clarifications from #72, #74, and #85 rebased for the new specification format.

It might be worth adding an ADR/other documentation about the possible methods for fast forward attack recovery and why this one was chosen.

cc @lukpueh @joshuagl @trishankatdatadog


joshuagl:
Thank you for working on this Marina! FYI I'm planning to review this early next week.

tuf-spec.md
```diff
@@ -1318,15 +1318,34 @@ it in the next step.
...
+  version.  To recover from a fast-forward attacks after the repository has been
+  compromised and recovered, certain metadata files need to be deleted as
+  specified in this section. If a delegated targets file is subjected to a
+  fast-forward attack, the snapshot role's keys should be replaced. Please see
+  [the Mercury paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
+  for more details on fast-forward attacks.
+
+    1. **Targets recovery** If a threshold of targets keys have been
+    removed in the new trusted root metadata compared to the previous trusted
+    root metadata, delete the old top-level targets and snapshot metadata
```
trishankatdatadog, line 1335: Don't you also want to delete timestamp?

tuf-spec.md
```diff
@@ -1318,15 +1318,34 @@ it in the next step.
...
+    files. Note that delegated targets are susceptible to fast forward attacks,
```
trishankatdatadog, line 1336: I have to admit these two sentences are not immediately clear to me. Would you please rephrase and clarify?

tuf-spec.md
```diff
@@ -1398,10 +1417,9 @@ it in the next step.
   in the trusted timestamp metadata.  If the versions do not match, discard the
   new snapshot metadata, abort the update cycle, and report the failure.
 
-5. **Check for a rollback attack**. The version number of the targets
-  metadata file, and all delegated targets metadata files, if any, in the
-  trusted snapshot metadata file, if any, MUST be less than or equal to its
-  version number in the new snapshot metadata file. Furthermore, any targets
+5. **Check for a rollback attack**. The version number of all targets metadata files in the
```
trishankatdatadog, line 1420: Nit: looks like line overflow

tuf-spec.md
```diff
@@ -1470,18 +1486,53 @@ it in the next step.
   2. Otherwise, recursively search the list of delegations in
      order of appearance.
 
-    1. If the current delegation is a multi-role delegation,
+    1. Let DELEGATEE denote the current target role TARGETS is
```
trishankatdatadog, line 1489: Could we please say DELEGATOR instead of TARGETS? Can be conflated at first glance with the top-level targets role.

tuf-spec.md
mnm678: As long as timestamp wasn't also subject to a fast forward attack, the timestamp shouldn't need to be deleted as the snapshot version number listed in timestamp remains valid. A new timestamp that signs the new snapshot will be released by the repository, and should be verified as usual.

tuf-spec.md
mnm678: This is using the same `TARGETS` defined at the beginning of step 7 (line 1475 in this version). I'll see if I can clarify this.

mnm678:
ping @trishankatdatadog @joshuagl

tuf-spec.md
trishankatdatadog: Couldn't the attacker fast-forward the timestamp to INT_MAX, too?

tuf-spec.md
trishankatdatadog: Sorry, just to be sparkling clear, I think we should start off by setting the DELEGATOR role to the top-level targets role, and go from there. Otherwise, it could confuse a few readers.

tuf-spec.md
mnm678: Yes, but that should be covered in 11.3. If the timestamp was the subject of a fast-forward attack, then the timestamp keys should also be replaced, triggering the recovery in 11.3.

tuf-spec.md
mnm678: better?

tuf-spec.md
```diff
@@ -1455,10 +1471,9 @@ it in the next step.
 6. **Persist targets metadata**. The client MUST write the file to
   non-volatile storage as FILENAME.EXT (e.g. targets.json).
 
-7. **Perform a pre-order depth-first search for metadata about the
-  desired target, beginning with the top-level targets role.** Note: If
-  any metadata requested in steps 5.6.7.1 - 5.6.7.2 cannot be downloaded nor
-  validated, end the search and report that the target cannot be found.
+7. **Perform a preorder depth-first search for metadata about the
+  desired target.** Let TARGETS be the current metadata, beginning with the
```
trishankatdatadog, line 1475:
```suggestion
  desired target.** Let DELEGATOR refer to the current
```

tuf-spec.md
```diff
@@ -1470,18 +1485,53 @@ it in the next step.
   2. Otherwise, recursively search the list of delegations in
      order of appearance.
 
-    1. If the current delegation is a multi-role delegation,
+    1. Let DELEGATOR be the current target role TARGETS and DELEGATEE denote
```
trishankatdatadog, line 1488:
```suggestion
    1. Let DELEGATEE denote
```

tuf-spec.md
```diff
@@ -1470,18 +1485,53 @@ it in the next step.
...
+    7. If the current delegation is a terminating delegation,
        then jump to step [[#fetch-target]].
 
-    3. Otherwise, if the current delegation is a
-       non-terminating delegation, continue processing the next delegation, if
-       any. Stop the search, and jump to step [[#fetch-target]] as soon as a delegation
-       returns a result.
+    8. Otherwise, if the current delegation is a non-terminating
+       delegation, continue processing the next delegation, if any, by repeating
+       the preorder depth-first search with DELEGATEE as the current TARGET role.
```
trishankatdatadog, line 1532:
```suggestion
       the preorder depth-first search with the DELEGATEE as the DELEGATOR.
```

tuf-spec.md
trishankatdatadog: I think it could be clearer. Please see my latest suggestions?

mnm678: That makes sense. I think yours maps more closely to what the code would look like.

tuf-spec.md
```diff
@@ -1470,18 +1485,53 @@ it in the next step.
...
        then jump to step [[#fetch-target]].
 
-    3. Otherwise, if the current delegation is a
-       non-terminating delegation, continue processing the next delegation, if
-       any. Stop the search, and jump to step [[#fetch-target]] as soon as a delegation
-       returns a result.
+    8. Otherwise, if the current delegation is a non-terminating
+       delegation, continue processing the next delegation, if any, by repeating
+       the preorder depth-first search with the DELEGATEE as the DELEGATOR.
+       Stop the search, and jump to step [[#fetch-target]]. as soon as a
```
trishankatdatadog, line 1533:
```suggestion
       Stop the search, and jump to step [[#fetch-target]] as soon as a
```

mnm678:
(I'll squash these last few commits once we finalize)

joshuagl:
> (I'll squash these last few commits once we finalize)

Possibly better to do it now, as squash & force push will discard any approvals.

mnm678:
> Possibly better to do it now, as squash & force push will discard any approvals.

Fair enough, it's either discard the approvals or discard the comment history :). I'll go ahead and fix it now

trishankatdatadog:
The whole preorder DFS is really not defined accurately. I can help to edit the text, but this [code](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/repository.py#L319-L403) might help in the meantime.

mnm678:
> The whole preorder DFS is really not defined accurately. I can help to edit the text, but this [code](https://github.com/trishankatdatadog/tuf-on-a-plane/blob/ee8049a63ea439d72509f6e876eef08fb50c6ab1/src/tuf_on_a_plane/repository.py#L319-L403) might help in the meantime.

Which part? This pr just adds the snapshot and hash checks for delegated targets

trishankatdatadog:
> Which part? This pr just adds the snapshot and hash checks for delegated targets

Yeah, maybe best done in another PR, but if you read the whole thing, you can see it has some work to do.

mnm678:
> Yeah, maybe best done in another PR, but if you read the whole thing, you can see it has some work to do.

I think I see what you mean. I'll move the check added in this pr to after the section about multi-role delegations, because these checks should be done for each role in the multi-role delegation. There might be some better clarifications to the section as a whole, but I'll leave that to a separate pr to avoid too much scope creep here.

tuf-spec.md
```diff
@@ -1471,17 +1486,53 @@ it in the next step.
...
+       does not match, abort the update cycle, and report the failure.
+
+    5. **Check for an arbitrary software attack.** The new DELEGATEE
+       metadata file MUST have been signed by a threshold of keys specified in the
+       DELEGATOR metadata file.  If the new DELEGATEE metadata file is not signed
+       as required, abort the update cycle, and report the failure.
+
+    6. **Check for a freeze attack.** The latest known time
+       should be lower than the expiration timestamp in the new DELEGATEE
+       metadata file. If so, the new DELEGATEE file becomes the trusted DELEGATEE
```
joshuagl, line 1524: We changed the way we describe a freeze attack check, could you make this consistent with the other checks in the spec?

"The expiration timestamp in new DELEGATEE metadata file MUST be higher than the fixed update start time."

joshuagl (changes requested):
This looks good. Some minor nits and the version & date will need to be bumped following the recent merge.

mnm678:
Thanks @joshuagl, I updated and rebased over the recent merge.

mnm678:
cc @joshuagl I rebased and updated the version number

tuf-spec.md
```diff
@@ -1479,13 +1494,55 @@ it in the next step.
...
+       to download the targets metadata file is of the fixed form FILENAME.EXT
+       (e.g., delegated_role.json).  Otherwise, the filename is of the form
+       VERSION_NUMBER.FILENAME.EXT (e.g., 42.delegated_role.json), where
+       VERSION_NUMBER is the version number of the DELEGATEE metadata file listed
+       in the snapshot metadata file.  In either case, the client MUST write the
+       file to non-volatile storage as FILENAME.EXT.
+
+    4. **Check against snapshot metadata.** The hashes (if any), and
+       version number of the new DELEGATEE metadata file MUST match the trusted
+       snapshot metadata, if any.  This is done, in part, to prevent a mix-and-match
```
joshuagl, line 1523: the "if any" has confused me, shouldn't we always have a trusted snapshot metadata at this point? as we download snapshot metadata before targets.

joshuagl (commented):
Thanks for the continued effort here @mnm678. I think we have a couple of ambiguities to iron out, but this is looking very solid overall.

tuf-spec.md
joshuagl, line 1523:
```suggestion
    4. **Check against snapshot metadata.** The hashes (if any), and
       version number (if any) of the new DELEGATEE metadata file MUST match the
       trusted snapshot metadata.  This is done, in part, to prevent a mix-and-match
```
Is this clearer?

joshuagl: For the top-level targets file we check snapshot hash (if any) in 5.6.2, then check targets is signed by a threshold of keys in 5.6.3, and then check snapshot version (if any) in 5.6.4. Is there a reason we do not do the same here?

Also, just noticed that [5.6](https://theupdateframework.github.io/specification/v1.0.20/#update-targets).4 does not indicate that the version is optional (hence my confusion above).

trishankatdatadog:
Overall LGTM pending resolution of Joshua's comments. Thanks, Marina!

tuf-spec.md
mnm678: Hmm, I actually think the ordering here is (possibly) wrong because you'd have to parse the metadata check the version before verifying the signatures. Either way I'll update to match the top-level targets order.

mnm678:
Thanks for the reviews @joshuagl and @trishankatdatadog! I added a couple of commits to address your comments.

trishankatdatadog (approved):

tuf-spec.md
```diff
@@ -1479,13 +1494,50 @@ it in the next step.
   2. Otherwise, recursively search the list of delegations in
      order of appearance.
 
-    1. If the current delegation is a terminating delegation,
+    1. Abort the update cycle, and report the failure.
```
joshuagl, line 1497: This doesn't look right? That step 1 is "Abort the update cycle"

mnm678: Oops, it looks like I should have removed this with the TAP 3 part.

trishankatdatadog:
This language is still vague:

> If a threshold of the <ROLE> keys have been removed in the new trusted root metadata compared to the previous trusted root metadata...

Do we mean no combination of the old threshold of the old <ROLE> keys have been removed on the new root metadata?

Cc @hosseinsia

mnm678:
> This language is still vague:
> 
> > If a threshold of the  keys have been removed in the new trusted root metadata compared to the previous trusted root metadata...
> 
> Do we mean no combination of the old threshold of the old keys have been removed in the new root metadata?
> 
> Cc @hosseinsia

Yes, does "If a threshold of keys from the previous trusted root metadata have been removed in the new trusted root metadata..." make this more clear?

trishankatdatadog:
> Yes, does "If a threshold of keys from the previous trusted root metadata have been removed in the new trusted root metadata..." make this more clear?

No, I think it means "...if _any_ threshold of keys from the previous trusted root metadata have been removed in the new trusted root metadata..."

Anyone disagree?

hosseinsia:
let's use an example:
(1) If the target role keys are [a, b, c, d] and threshold is 2 in v1 and the new keys are [b, c, d, e] then no need to remove targets.json because only one key is being revoked and an attacker could not have run the fast-forward attack.

(2) If the target role keys are [a, b, c, d] and threshold is 2 in v1 and the new keys are [c, d, e] then we have to remove targets.json

The reasoning is we assume revoked keys are compromised and and attacker might have launched a fast-forward attack on targets.json because he had enough keys.

Do you agree?

So I think both of you are talking about the same thing?

mnm678:
> Do you agree?
> 
> So I think both of you are talking about the same thing?

I agree, @trishankatdatadog is this also what you were talking about? I think the biggest problem is making this clear in the wording of the spec. Maybe an example like the one above would help.

trishankatdatadog:
> > Do you agree?
> > So I think both of you are talking about the same thing?
> 
> I agree, @trishankatdatadog is this also what you were talking about? I think the biggest problem is making this clear in the wording of the spec. Maybe an example like the one above would help.

I think we are all in the same page now, which goes back to [this](https://github.com/theupdateframework/specification/pull/150#issuecomment-909417445) interpretation.

To remove any ambiguity, I propose we write how to compute whether the keys have been rotated by this interpretation, so there is no room for error.

hosseinsia:
We might as well need to clarify our assumptions about what we believe people should do when there is a key compromise for a role. Options:
(1) Revoke all the keys for that role and create new ones.
(2) Only revoke the keys that we think they are being compromised. 

With the first assumption, we remove the related top-level metadata when all of the keys in the previous version are revoked. This is because any other type of change does not signal key compromise.

With the second assumption, we remove the related top-level metadata when the count of revoked keys is equal or bigger than the threshold of the previous version.

Which one should we assume?

trishankatdatadog:
> (2) Only revoke the keys that we think they are being compromised.

We should assume that this is the only option people can take because they cannot afford to switch out the other keys.

hosseinsia:
ok then. Borrowing an example notations from Trishank to formulate the problem and answer:

**Question:** Let file A be signed by a threshold m_t of keys in k_t at time t
And A be signed by a threshold $m_{t+n}$ of keys in $k_{t+n}$ at time t+n
How do you tell whether the keys for $A_{t+n}$ have been rotated (i.e., not verifiable by keys for A_t) w/o opening $A_{t+n}$? 

**Note:** The key rotation signals the need for removing metadata for the role. 

**Answer:** Count the number of deleted keys , if it is bigger or equal to the the $m_t$ then the keys are rotated. In other words, if  $$k_t - k_{t+n} >= m_t$$ is true then the key is rotated.

WDYT?

trishankatdatadog:
> How do you tell whether the keys for $A_{t+n}$ have been rotated (i.e., not fully verifiable by any thresholds of keys for A_t) w/o opening $A_{t+n}$?

I would change this to:

> How do you tell whether any threshold $m_t$ of $k_t$ have been removed in $k_{t+n}$?

tuf-spec.md
```diff
@@ -1327,15 +1327,32 @@ it in the next step.
...
+  compromised and recovered, certain metadata files need to be deleted as
+  specified in this section. If a delegated targets file is subjected to a
+  fast-forward attack, the snapshot role's keys should be replaced. Please see
+  [the Mercury paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
+  for more details on fast-forward attacks.
+
+    1. **Targets recovery** If a threshold of targets keys have been
+    removed in the new trusted root metadata compared to the previous trusted
+    root metadata, delete the old top-level targets and snapshot metadata
+    files. Note that this only applies to top-level targets metadata whose
```
jku, line 1345: "if targets is no longer correctly signed, delete local snapshot"

why? this makes no sense to me. If a repository wants a new snapshot won't it just publish one? If a repository wants to make sure old snapshot is not used even for rollback protection, it can rotate the snapshot keys

tuf-spec.md
```diff
@@ -1327,15 +1327,32 @@ it in the next step.
...
+    keys are listed in root metadata.
+
+    2. **Snapshot recovery** If a threshold of snapshot keys have
+    been removed in the new trusted root metadata compared to the previous
+    trusted root metadata, delete the old snapshot and timestamp metadata
+    files.
```
jku, line 1351: why is the timestamp deleted? If timestamp is compromised or repository wants the local timestamp to not be used for rollback checks, it can rotate the timestamp key

jku (commented):
Can someone explain why there need to be these rules about which metadata needs to be deleted in which rotation cases... why is it not enough to say _"if any local metadata is not signed by current threshold of keys, that local metadata should be considered not valid and should not be used as trusted metadata"_? 

The rule could be spelled out for specific cases but still in the same format:
"If local timestamp is not signed by threshold of keys defined in the new/current root, local timestamp should be considered not valid and should not be used as trusted metadata"

This is what we care about in the end right -- is the metadata signed by current threshold of keys or not? Which keys were rotated and how many keys were rotated is not really relevant -- or have I misunderstood something?

trishankatdatadog:
> Can someone explain why there need to be these rules about which metadata needs to be deleted in which rotation cases... or have I misunderstood something?

Precisely for the latter reason ;) Not everyone who reads the text has the same idea, which is a bad thing for building consistent and correct implementations.

trishankatdatadog:
> What I mean is: I believe we only care whether _metadata is signed by the current threshold of keys_. It it is not, it should not be used as trusted metadata.

What do you mean by current? Would you use the definitions [here](https://github.com/theupdateframework/specification/pull/150#issuecomment-909789484) so we are all on the same page?

trishankatdatadog:
> > > What I mean is: I believe we only care whether _metadata is signed by the current threshold of keys_. It it is not, it should not be used as trusted metadata.
> > > What do you mean by current?
> 
> Whenever we load a delegator like root we get current keys/thresholds for the delegated metadata. If delegated metadata is not signed by these keys, it should not be trusted after this point. So in practice: if a new root is loaded with new timestamp keys, the local timestamp we might have on disk (or even loaded in memory) is not signed with those keys and cannot be trusted from now on.

So IIUC you mean if no m_t of k_t exists in m_{t+n} of k_{t+n}, then delete previous timestamp/snapshots/targets metadata.

This was also my interpretation, but Hossein pointed out that it's too strong. Look at his example [here](https://github.com/theupdateframework/specification/pull/150#issuecomment-909459074).

Just because a threshold of old keys have colluded and caused fast-forward attacks, doesn't mean you have the luxury to rotate all the rest that have not provably colluded.

So, the requirement should be less strict: if _any_ (not all) old threshold of old keys have been rotated out, then take that as a signal of possible fast-forward attack, and delete the previous metadata.

jku:
> So IIUC you mean if no m_t of k_t exists in m_{t+n} of k_{t+n}, then delete previous timestamp/snapshots/targets metadata.

I have to admit I don't understand what that means... but I'll look at the examples:

> (1) If the target role keys are [a, b, c, d] and threshold is 2 in v1 and the new keys are [b, c, d, e] then no need to remove targets.json because only one key is being revoked (for being compromised) and an attacker could not have run the fast-forward attack.

If we have loaded root v2 and current targets keys are [b,c,d,e], we absolutely cannot trust targets.json **if** it was signed by only two keys and one of them was a, because at that point _it is not signed by threshold of keys_.

On the other hand if targets.json was signed by more than two keys (or one of them was not a), things are just fine.

> (2) If the target role keys are [a, b, c, d] and threshold is 2 in v1 and the new keys are [c, d, e] then we have to remove targets.json

If targets.json was signed by at least c & d, then we can trust it and do not have to remove it :shrug: 

From the context I finally think I understand what you are getting at: you think you can use the number of removed keys in a specific delegator update in comparison to the threshold of keys as a signal... I don't think you can as the keys might have been removed one by one in multiple metadata versions -- the possibility of a ffwd attack is still there even though only one key was removed at a time. 

I don't quite understand why you want to remove targets.json if you suspect a ffwd attack may have happened? the real issue is in snapshot (where the targets.json version might be too big) and the issue may have been there for multiple snapshot versions already (so you can't just go back to a previous snapshot or something). I don't think there are other solutions to this problem than snapshot key rotation, and frankly it seems like a really good solution to me...

trishankatdatadog:
> we absolutely cannot trust targets.json **if** it was signed by only two keys and one of them was a, because at that point _it is not signed by threshold of keys_.

The key problem is that we are looking at only the previous and current roots now to decide whether to delete previous timestamp/snapshot/targets metadata _without_ looking at the current version of the latter, you see? So you cannot check based on what the actual signatures are. You can only decide based on comparing previous/current threshold of keys.

mnm678:
As I see it there are 2 things being discussed (and possibly confused) here:

1. Do we need explicit fast forward attack workflows in the client for non-snapshot metadata? If targets or timestamp are fast-forwarded, this can be solved by replacing the snapshot keys on the repo to trigger the snapshot metadata workflow.
2. When doing the client fast forward attack recovery, when should this recovery be triggered?

For 1, this logic is similar to the reason for not providing [explicit fast-forward attack recovery for delegated targets](https://github.com/theupdateframework/specification/pull/86#issuecomment-6887254030). I'm starting to agree with @jku that snapshot might be the only place that fast forward attack recovery is needed.

2 is a bit trickier. If you have a previous trusted snapshot metadata file A and a new one B (both signed by a valid threshold of keys from the root metadata at the time they were downloaded), how do you know if you should do rollback checks between B and A, or skip these to recover from a possible fast forward attack. I would argue that the rollback checks should be skipped whenever a threshold<sub>A</sub> of keys from A  is not present in B (which I think is what @trishankatdatadog is arguing).

Am I missing anything?

jku:
> > we absolutely cannot trust targets.json **if** it was signed by only two keys and one of them was a, because at that point _it is not signed by threshold of keys_.
> 
> The key problem is that we are looking at only the previous and current roots now to decide whether to delete previous timestamp/snapshot/targets metadata _without_ looking at the current version of the latter, you see? 

This is what I'm suggesting, no? If the new root rotated targets keys in a way that the local targets.json is no longer signed by correct threshold of keys, we should not trust that targets.json for anything. I'm not waiting for some next version of targets before making that decision, I'm deciding that immediately as the keys are rotated.

> You can only decide based on comparing previous/current threshold of timestamp/snapshot/targets keys.

I agree that we should only look at the keys/threshold, but I'm arguing we can only consider the currently valid keys/threshold. The idea that we can somehow compare current and previous keys is just not sound. Just as one example, imagine this scenario:
* client updates root from remote, gets a new version N+1 that includes key rotations
* update process gets somehow canceled before other metadata is updated. N+1 is now the current root version
* client starts the update again, updates root to N+2 (it has no additional key rotations)
* at this point N+1 is already the "previous" version, so the  ffwd detection does not trigger

jku:
> 2 is a bit trickier. If you have a previous trusted snapshot metadata file A and a new one B (both signed by a valid threshold of keys from the root metadata at the time they were downloaded), how do you know if you should do rollback checks between B and A, or skip these to recover from a possible fast forward attack. I would argue that the rollback checks should be skipped whenever a thresholdA of keys from A is not present in B (which I think is what @trishankatdatadog is arguing).

I still don't understand why the repository can't handle this. If there is a fast forward attack I think it's completely reasonable to expect the repository to update root and rotate the keys of the snapshot metadata that we do not want to rollback check: key rotation logically leads to client not using the old snapshot for the rollback checks (as at that point the old snapshot is _not_ signed by current threshold of snapshot keys).  is this for some reason a bad solution?

mnm678:
> I still don't understand why the repository can't handle this. If there is a fast forward attack I think it's completely reasonable to expect the repository to update root and rotate the keys of the snapshot role: key rotation logically leads to client not using the old snapshot for the rollback checks (as at that point the old snapshot is _not_ signed by current threshold of snapshot keys). is this for some reason a bad solution?

Oh, I think I understand what you're saying now. The client would use the keys in the current root file to verify both the previous and the new snapshot, then if the verification for the previous failed with the current keys, the rollback check would be skipped. Is that what you're talking about? That does seem simpler than the previous/current comparison

jku:
> > I still don't understand why the repository can't handle this. If there is a fast forward attack I think it's completely reasonable to expect the repository to update root and rotate the keys of the snapshot role: key rotation logically leads to client not using the old snapshot for the rollback checks (as at that point the old snapshot is _not_ signed by current threshold of snapshot keys). is this for some reason a bad solution?
> 
> Oh, I think I understand what you're saying now. The client would use the keys in the current root file to verify both the previous and the new snapshot, then if the verification for the previous failed with the current keys, the rollback check would be skipped. Is that what you're talking about? That does seem simpler than the previous/current comparison

Yes! I think it's a fundamental idea that's already encoded into every rule in the client workflow but for some reason not spelled out: **do not trust or use metadata that is not signed by _current_ keys for that role**. This should apply at all times: If the local ("old") snapshot metadata is not signed by _current_ snapshot keys after a root update, the "old" snapshot should not be used for rollback checks (or anything at all, because it's not trusted!).

trishankatdatadog:
> Yes! I think it's a fundamental idea that's already encoded into every rule in the client workflow but for some reason not spelled out: **do not trust or use metadata that is not signed by _current_ keys for that role**. This should apply at all times: If the local ("old") snapshot metadata is not signed by _current_ snapshot keys after a root update, the "old" snapshot should not be used for rollback checks (or anything at all, because it's not trusted!).

We still disagree if I understand you correctly.

Take Hossein's example of m_t = 2, k_t = [a,b,c,d], m_{t+n} = 2, k_{t+n} = [c,d,e].

You can still use m_{t+n} and k_{t+n} (specifically [c,d]) to verify A_t, but that doesn't mean [a,b] wasn't used to fast-forward snapshot (fuhgedabout targets, no need to get distracted by that red herring). Hence Hossein's condition: if he sees _any_ m_t of k_t removed in k_{t+n}, he reasonably and automatically assumes a fast-forward.

trishankatdatadog:
Really, the best resolution is to use an explicit instead of implicit, ambiguous signal to delete timestamp/snapshot/targets metadata from one root to another, but that requires a spec change. Might be worth it than trying to get consensus by abusing an existing, orthogonal mechanism.

jku:
> Really, the best resolution is to use an explicit instead of implicit, ambiguous signal to delete timestamp/snapshot/targets metadata from one root to another, but that requires a spec change. Might be worth it than trying to get consensus by abusing an existing, orthogonal mechanism.

rotating keys is an explicit unambiguous signal that the old metadata (now no longer signed by current keys) should not be trusted and should not be used for anything, including rollback checks.  I don't see how it's an orthogonal mechanism: this property is the signatures sole reason for existence. Do we really not have consensus on that?

jku:
> I agree, but I don't think 'key rotation' is currently well defined in the spec.

yes, apologies: I meant any change to the roles keys/threshold that leads to local/old metadata to be not trusted.

mnm678:
@trishankatdatadog @jku I updated the fast-forward attack recovery section. Does this look good to both of you?

tuf-spec.md
```diff
@@ -1334,26 +1334,16 @@ it in the next step.
...
-    1. **Targets recovery** If a threshold of targets keys have been
-    removed in the new trusted root metadata compared to the previous trusted
-    root metadata, delete the old top-level targets and snapshot metadata
-    files. Note that this only applies to top-level targets metadata whose
-    keys are listed in root metadata.
-
-    2. **Snapshot recovery** If a threshold of snapshot keys have
-    been removed in the new trusted root metadata compared to the previous
-    trusted root metadata, delete the old snapshot and timestamp metadata
+    1. **Snapshot recovery** If the trusted snapshot metadata cannot be
```
trishankatdatadog, line 1342: @hosseinsia thoughts?

tuf-spec.md
```diff
@@ -1334,26 +1334,16 @@ it in the next step.
...
+    validating using a threshold of snapshot keys from the new trusted root
```
hosseinsia, line 1343:
```suggestion
    validated using a threshold of snapshot keys from the new trusted root
```

tuf-spec.md
hosseinsia: This is clever. I like it! So at least a combination of non-revoked keys should have signed the metadata?
And an attacker who uses the compromised keys to sign metadata won't succeed.

tuf-spec.md
```diff
@@ -1334,15 +1334,22 @@ it in the next step.
...
+  compromised and recovered, certain metadata files need to be deleted as
+  specified in this section. If a targets file is subjected to a
+  fast-forward attack, the snapshot role's keys should be replaced. Please see
+  [the Mercury paper](https://ssl.engineering.nyu.edu/papers/kuppusamy-mercury-usenix-2017.pdf)
+  for more details on fast-forward attacks.
+
+    1. **Snapshot recovery** If the trusted snapshot metadata cannot be
+    validated using a threshold of snapshot keys from the new trusted root
+    metadata, delete the trusted snapshot and timestamp metadata
+    files.
```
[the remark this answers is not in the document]
jku: In fact, how would this heuristic even work in practice:
* client has a timestamp with snapshot version==1billion
* client starts a refresh, downloads root, then downloads new timestamp with snapshot version==1
* client sees a rollback attack, stops the update and never even downloads the snapshot that would include the heuristic that would make the valid timestamp now invalid

jku (commented):
I think we're moving closer to an agreement but making timestamp untrusted because of something in snapshot changes (even if it's the snapshot keys) still sounds fundamentally wrong.

Left a longer comment inline but that's the core issue I have I think.

tuf-spec.md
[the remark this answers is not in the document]
mnm678: I agree that we should separate back out the timestamp vs snapshot recovery. Looking at it again, I actually don't think this check is in the right place in the workflow. I think the timestamp recovery should happen when the timestamp is downloaded, not as part of the root metadata workflow, and the same for snapshot. Especially as it is no longer tied to anything in the previous root metadata file.

It would be great to have separate documentation for the repository workflow. Right now all the advise for managing a fast forward attack (and a lot of other pieces) is buried here in the client workflow.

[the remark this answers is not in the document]
trishankatdatadog: Friends, but this is where and what we recommend in [the Mercury paper](https://www.usenix.org/conference/atc17/technical-sessions/presentation/kuppusamy). See "5.3 Recovering from a repository compromise" on page 7.

This is getting contentious enough (for good reasons) that I'm not sure that endless cycles of reviews and comments would do sufficient justice. A meeting hasn't solved the problem either. Perhaps it's time for the BDFL to step in and make a decision one way or another? @JustinCappos

JustinCappos:
I've re-read the thread and feel like we could perhaps do with another
meeting about this.  I can unblock us if we can't resolve it after another
round of discussion.

I did spend a fair amount of time thinking about a way to explicitly
indicate that the metadata should be deleted, but could not think of a good
way to do so when considering clients that may miss updates over an
extended period.

trishankatdatadog:
> I've re-read the thread and feel like we could perhaps do with another meeting about this. I can unblock us if we can't resolve it after another round of discussion.

Sure, I am willing to do one more meeting.

> I did spend a fair amount of time thinking about a way to explicitly indicate that the metadata should be deleted, but could not think of a good way to do so when considering clients that may miss updates over an extended period.

Why not just set an explicit signal in the root metadata? You will always pick it up when you rotate roots.

JustinCappos:
If you set the signal, do you remove it when you next rotate roots
normally?  What if a client misses the intermediate update?  What if there
is also a spec version upgrade?  Does a client always get all root files?

joshuagl (commented):
Thanks for all the great discussion here, apologies for missing most of it. 

I'd be willing to do have another discussion, but I think that might be more productive if we capture the options being proposed as a prerequisite?

jku:
Just a related note here: we now have tests for all rollback attacks and all fast-forward recovery cases in python-tuf ngclient thanks to @kairoaraujo (search for "rollback" in https://github.com/theupdateframework/python-tuf/blob/develop/tests/test_updater_top_level_update.py)

No ff-recovery specific rules are needed in the client. The only rule is:
**only ever use metadata that is signed with current threshold of current keys, do not trust any other metadata**

This directly leads to following matrix because of the rollback checks:
```
| to rollback | these keys must be rotated |
|-------------+----------------------------|
| timestamp   | timestamp                  |
| snapshot    | timestamp, snapshot        |
| targets     | snapshot                   |
```
(where "rotate" means that no previous root should contain keys such that the signatures on new rolled back metadata would be valid).

I believe that is in line with this PR and I think the PR is an improvement... but I think long term the spec would be more useful to implementers if it was built around this idea (_only ever use metadata that is signed with current threshold of current keys_) and not specific individual rules (like _if specific situation A then delete files B and C_): The same advice can still be given but it would be easier to understand cause-and-effect and design good software.

PR #152, opened 2021-03-24
Update README links to the rendered specification

sechkova:
Point to the bikeshed generated specification from the "latest" and "draft" links in README.

Signed-off-by: Teodora Sechkova <tsechkova@vmware.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,8 @@
 The Update Framework specification
 ----------------------------------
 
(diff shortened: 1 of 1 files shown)


lukpueh (approved):
Links resolve as they should. LGTM. Thanks, Teodora! 

Given that this PR doesn't change the spec, I'll merge without passing CI/CD or second review...

Issue #105, opened 2020-07-06
Describe how updates should occur on the repository side


joshuagl:
> the spec does not have a notion of developer uploads. As a matter of fact, and I think this should be changed, right now it does not even describe how updates occur on the repository-side (as we do in [PEP-458](https://github.com/python/peps/blob/master/pep-0458.txt#L1008-L1019)).

_Originally posted by @lukpueh in https://github.com/theupdateframework/specification/issues/104#issuecomment-651073233_

joshuagl:
Related issue in the reference implementation https://github.com/theupdateframework/tuf/issues/969

joshuagl:
I've opened a PR #126 to include the recommended metadata write order, as detailed in PEP 458.

I think it makes sense to include more detailed workflows/recommendations in the proposed secondary literature (#91).

joshuagl:
Resolved by #153

PR #153, opened 2021-03-29
Consolidate repository related content, describe repository workflow for adding targets

joshuagl:
This PR renames the "Usage" section to "Repository operations" and moves the section on "Consistent snapshots" to be a subsection of "Repository operations" alongside the existing subsection on "Key management and migration".

Then, most notable, in 4ccd04c it adds a new subsection to "Repository operations" to describe the workflow for updating repository metadata when adding targets to the repository, or updating existing targets on the repository. This lifts heavily from the [text in PEP 458](https://github.com/python/peps/blob/master/pep-0458.txt#L1008-L1019) describing the repository workflow, fixing #105.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1321,7 +1321,7 @@ it in the next step.
 11. **If the timestamp and / or snapshot keys have been rotated, then delete the
   trusted timestamp and snapshot metadata files.** This is done
   in order to recover from fast-forward attacks after the repository has been
-  compromised and recovered. A _fast-forward attack_ happens when attackers
+  compromised and recovered. A *fast-forward attack* happens when attackers
   arbitrarily increase the version numbers of: (1) the timestamp metadata, (2)
   the snapshot metadata, and / or (3) the targets, or a delegated targets,
   metadata file in the snapshot metadata. Please see [the Mercury
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1620,6 +1620,51 @@ so that outdated clients can update to the latest available root.
 
 See [[#detailed-client-workflow]] for more details.
 
+## Adding and updating targets ## {#adding-updating-targets}
+
+The following subsections describe how to update metadata on the repository
+when adding targets to the repository, or updating existing targets.
+
+### Update targets metadata ### {#update-targets-metadata}
+
+1. Add the new (or update an existing) <a>TARGETS</a> object in the relevant
+   targets metadata (either the top-level targets metadata, or a delegated
+   targets metadata).
+2. Increment the <a for="role">VERSION</a> number in the updated targets
+   metadata.
+3. Sign the updated targets metadata with at least a <a>THRESHOLD</a> of keys
+   for the associated targets role (either the top-level targets role, or a
+   delegated targets role).
+4. Write the updated targets metadata, ensuring the targets metadata filename is
+   prefixed with the <a for="role">VERSION</a> number if consistent snapshots
+   are enabled for the repository.
+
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1505,10 +1505,10 @@ it in the next step.
    metadata file found earlier in step [[#update-targets]].  In either
    case, the client MUST write the file to non-volatile storage as FILENAME.EXT.
 
-# 6. Usage # {#usage}
+# Repository operations # {#repository-operations}
```
lukpueh, line 1508: Not directly related to this PR, but would you mind changing:
- `section 6.1` to `Section [[#key-management-and-migration]]` in [tuf-spec.md#L307](https://github.com/theupdateframework/specification/blame/6fc41b93a4a31914348b8b2baaa5d2cf126c5644/tuf-spec.md#L307), and 
- `Section 7` to `Section [[#consistent-snapshots]]` in [tuf-spec.md#L692](https://github.com/theupdateframework/specification/blame/6fc41b93a4a31914348b8b2baaa5d2cf126c5644/tuf-spec.md#L692)
- ?

tuf-spec.md
```diff
@@ -1616,10 +1616,53 @@ released versions of root metadata files should always be provided
...
+when adding targets to the repository, or updating existing targets.
+
+### Update targets metadata ### {#update-targets-metadata}
+
+1. Add the new (or update an existing) <a>TARGETS</a> object in the relevant
+   targets metadata (either the top-level targets metadata, or a delegated
+   targets metadata).
+2. Increment the <a for="role">VERSION</a> number in the updated targets
+   metadata.
+3. Sign the updated targets metadata with a threshold of keys for the associated
```
lukpueh, line 1635: Is it worth to...
```suggestion
3. Sign the updated targets metadata with a <a>THRESHOLD</a> of keys for the associated
```
?

Also, should it be *"at least a threshold"*? Or is that implicit?

Same comment applies below.

lukpueh (changes requested):
Thanks a ton for the patch, @joshuagl. The new section will be very helpful for TUF novices. Please consider addressing some minor comments and we should be good to go.

tuf-spec.md
```diff
@@ -1616,10 +1616,53 @@ released versions of root metadata files should always be provided
...
 
 
-## Reading consistent snapshots ## {#reading-consistent-snapshots}
+### Reading consistent snapshots ### {#reading-consistent-snapshots}
 
 See [[#detailed-client-workflow]] for more details.
 
+## Adding and updating targets ## {#adding-updating-targets}
+
+The following sub-sections describe how to update metadata on the repository
```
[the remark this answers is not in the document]
joshuagl: Good catch, thanks!

tuf-spec.md
joshuagl: Nice. Done.

tuf-spec.md
joshuagl: I agree, linking to the definition seems worthwhile. I also think it's better to be explicit than implicit, so have expanded to "at least a threshold".

joshuagl:
Thanks for the review @lukpueh ! I've squashed the suggested changes into the existing patches, apologies if that makes the re-review harder.

lukpueh (dismissed):

joshuagl:
I think this should be non controversial. Do you have cycles to review please @mnm678 & @trishankatdatadog ?

This got one review from @lukpueh which was dismissed when I pushed an extra change.

trishankatdatadog (approved):
About time, thank you!

mnm678 (approved):
LGTM!

PR #157, opened 2021-04-21
Clarify that delegations are optional

MVrachev:
Nowhere in the spec, we clarify that "delegations" is an optional field
in the targets metadata file.
This is a possible reason why (at the time of writing this commit) in
the TUF python reference implementation "delegations" is still a
required field.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2020-12-11
+Date: 2021-04-21
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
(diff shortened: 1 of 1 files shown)


mnm678 (approved):
Great clarification. I believe this preserves backwards compatibility (required -> optional), so LGTM.

jku:
> the TUF python reference implementation "delegations" is still a required field.

What is this based on btw (where is it required)?

MVrachev:
In the new metadata API:
https://github.com/theupdateframework/tuf/blob/41f7e809fef4cfe578cd2bd96f497da74b1bce15/tuf/api/metadata.py#L690

jku:
Ah right: so it's a bug in new API. Fixing that definitely does not preserve backwards compatibility (a client that assumed delegations are required bys using the current API _will_ break if the server now  makes delegations optional) but that seems ok if it's in api/metadata.py...

In any case that's an implementation issue: the spec is not changing its position, it's clarifying an existing one as I see it: the definition of "signed" already has delegations in parentheses (the use of parenthesis in the spec does not seem entirely consistent but in this case I read that as meaning "optional") . LGTM.

joshuagl:
Per the [scope](https://theupdateframework.github.io/specification/v1.0.18/index.html#scope) use keywords from [RFC 2119](https://tools.ietf.org/html/rfc2119), so we should say OPTIONAL in uppercase. Custom is also optional, could you add a change to address that also?

MVrachev:
Superseded by #165.

Issue #156, opened 2021-04-20
Clarify "paths" and "path_hash_prefixes" in delegations


MVrachev:
In the specification [here](https://theupdateframework.github.io/specification/latest/#path_hash_prefixes), it's said:
```
In order to discuss target paths, a role MUST specify only one of the
"path_hash_prefixes" or "paths" attributes, each of which we discuss next.
```

This is confusing because it doesn't properly define which of these options is valid:
- BOTH `paths` and `path_hash_prefixes`
- ONLY ONE of `paths` and `path_hash_prefixes`
- NONE of `paths` and `path_hash_prefixes`

joshuagl:
Thank you for filing, we can definitely work on making this clearer. A valid DELEGATIONS object can have one or none of `paths` or `path_hash_prefixes` (bullets two and three from your list).

jku:
> valid DELEGATIONS object can have one or none of `paths` or `path_hash_prefixes` (bullets two and three from your list).

I agree this seems logical but will also point out that the current/legacy updater will currently raise an error if neither is defined.

PR #158, opened 2021-04-21
Clarify "paths" and "path_hash_prefixes"

MVrachev:
**Note!** 
The `Text Macro: VERSION` is bumped by two with the idea of merging https://github.com/theupdateframework/specification/pull/157 before this pr.

Fixes: https://github.com/theupdateframework/specification/issues/156

Clarify "paths" and "path_hash_prefixes" in delegations, because
currently, it's not properly defined which of these options can be used
to create a valid target file:
- BOTH paths and path_hash_prefixes
- ONLY ONE of paths and path_hash_prefixes
- NONE of paths and path_hash_prefixes

With this change, I aim to define clearly that a valid target file will
contain ONLY ONE of them or NONE of them.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2020-12-11
+Date: 2020-04-21
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1018,9 +1018,8 @@ as is described for the <a>root.json</a> file.
     package that are not made by the delegated party or its descendants to be
     ignored.
 
-In order to discuss target paths, a role MUST specify only one of the
-<a>"path_hash_prefixes"</a> or <a for="delegation-role">"paths"</a> attributes, each of which we
-discuss next.
+The <a>"path_hash_prefixes"</a> and <a for="delegation-role">"paths"</a>
+attributes are optional, if used, exactly one of them should be set.
```
joshuagl, line 1022:
```suggestion
attributes are OPTIONAL, if used, exactly one of them should be set.
```
We use RFC 2119 keywords in the spec.

MVrachev:
Superseded by #165.

PR #161, opened 2021-05-12
Add dependabot config

joshuagl:
Adds a dependabot config to monitor for updates in the GitHub Actions we are using.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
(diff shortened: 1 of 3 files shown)


mnm678 (approved):
The CI failure is due to the date check, and the actual change ran fine, so LGTM. Is it possible to skip the date/version check for non-spec changes?

joshuagl:
> The CI failure is due to the date check, and the actual change ran fine, so LGTM. Is it possible to skip the date/version check for non-spec changes?

It is, I've added 2fa4c01 to change check_release.py to only check version and date when _tuf-spec.md_ is in the list of modified files.

mnm678 (approved):
Thanks for fixing the ✅

joshuagl:
@trishankatdatadog could you please take a look at this PR when you have a chance?

trishankatdatadog (approved):

PR #162, opened 2021-05-17
Clarify that CONSISTENT_SNAPSHOT is optional

MVrachev:
From chapter 7 in the spec (version 1.0.17):
"Finally, the root metadata should write the Boolean
"consistent_snapshot" attribute at the root level of its keys of
attributes.
If consistent snapshots are not written by the repository,
then the attribute may either be left unspecified or be set to the
False value. Otherwise, it must be set to the True value."

The above implies that there could be repositories with root metadata
without CONSISTENT_SNAPSHOT.
Clarify that, but phrase it to make it clear this should be included
in new implementations.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2020-12-11
+Date: 2021-05-17
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.17
+Text Macro: VERSION 1.0.18
 </pre>
 
 Note: We strive to make the specification easy to implement, so if you come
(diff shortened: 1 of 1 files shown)


jku (commented):
Other descriptions of  optional fields use the word "OPTIONAL" in capitals, maybe you want to do that too. Also, it seems the metadata examples use ( ) to signify optional parts so maybe we should have  `("consistent_snapshot": <a>CONSISTENT_SNAPSHOT</a>,)` in there as well.

There's an extra d in "include"

tuf-spec.md
```diff
@@ -688,9 +688,11 @@ The "signed" portion of <a>root.json</a> is as follows:
 
   : <dfn>CONSISTENT_SNAPSHOT</dfn>
   ::
-    A boolean indicating whether the repository supports
-    consistent snapshots.  Section 7 goes into more detail on the consequences
-    of enabling this setting on a repository.
+    An optional boolean indicating whether the repository supports
+    consistent snapshots. This field is optional for backward compatibility with
+    old metadata that may exist without it. New implementations SHOULD included
```
trishankatdatadog, line 693:
```suggestion
    old metadata that may exist without it. New implementations SHOULD include
```

trishankatdatadog, line 693: Where do we discuss what we should do if it doesn't exist?

MVrachev: The spec leaves this open. From [chapter 7.1](https://theupdateframework.github.io/specification/latest/#writing-consistent-snapshots):
```
Finally, the root metadata should write the Boolean "consistent_snapshot" 
attribute at the root level of its keys of attributes. 
If consistent snapshots are not written by the repository, 
then the attribute may either be left unspecified or be set to the False value. 
Otherwise, it must be set to the True value.
```

tuf-spec.md
```diff
@@ -688,9 +688,11 @@ The "signed" portion of <a>root.json</a> is as follows:
...
   : <dfn>CONSISTENT_SNAPSHOT</dfn>
   ::
-    A boolean indicating whether the repository supports
-    consistent snapshots.  Section 7 goes into more detail on the consequences
-    of enabling this setting on a repository.
+    An optional boolean indicating whether the repository supports
+    consistent snapshots. This field is optional for backward compatibility with
+    old metadata that may exist without it. New implementations SHOULD include
+    it. Section 7 goes into more detail on the consequences of enabling
+    this setting on a repository.
```
joshuagl, line 695: Let's both ensure we are cross-linking and avoid referring to sections by their explicit numbering
```suggestion
    it. [[consistent-snapshots]] goes into more detail on the consequences of enabling
    this setting on a repository.
```

joshuagl (changes requested):
Thank you for suggesting this. A few minor comments. We'll also need to address the version number for this and your other submissions. Probably makes sense to come up with a planned merge order with the other maintainers – let me get back to you on that.

MVrachev:
Superseded by #165.

PR #166, opened 2021-06-02
Add notes on uniqueness of keyids and role names

jku:
I'm making explicit some requirements that I _think_ are already implicit in the spec:
* role names in delegations must be unique
* keyids in signatures must be unique

I guess the high level question is: is this really a "bug fix" or should it be handled as a change in spec? I think it's a fix that clarifies what is already the obvious/logical/safe implementation of the spec, but especially on the role names I can see someone could maybe argue otherwise.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1004,6 +1004,8 @@ as is described for the <a>root.json</a> file.
   : <dfn>ROLENAME</dfn>
   ::
     A string giving the name of the delegated role.  For example, "projects".
+    The rolename MUST be unique in the delegations object: multiple roles with
+    the same rolename are not allowed within a <dfn>DELEGATIONS</dfn>.
 
   : <dfn>TERMINATING</dfn>
   ::
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -529,6 +529,8 @@ All signed metadata objects have the format:
       ::
         The identifier of the key signing the <a for="role">ROLE</a> object,
         which is a hexdigest of the SHA-256 hash of the canonical form of the key.
+        The keyid MUST be unique in the "signatures" array: multiple
+        signatures with the same keyid are not allowed.
 
       : <dfn>SIGNATURE</dfn>
       ::
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-05-27
+Date: 2021-06-02
 Editor: Justin Cappos, NYU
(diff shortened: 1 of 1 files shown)


mnm678:
The challenge for both of these requirements is what scope they need to be unique within. I agree that within a single targets metadata file, it would be wrong to have non-unique keyids, but what about between two different delegated targets metadata files? There might be a reason to have the same keyid mean different things in different role's delegations.

But with a well-defined scope, I think that this could be a useful clarification to the specification.

jku:
Yes, agreed on the future goal: uniqueness of keys should mean uniqueness within the "signatures" object only. 

But as a bug fix I think this is the best we can do -- leave the keyid definition as is but clarify the aspect that makes safer implementations possible (an aspect that is already implicit: if the keyid is a hash it should be unique). 

Admittedly, delegated role uniqueness is not really even implicit in the spec, it's just that I think it's what the authors intended because otherwise the delegation verification is a mess? The fact that roles in Targets is not a dictionary is just a side-effect of json not having an ordered dictionary... otherwise I expect it would be a dict just like root roles are.

jku:
> I think it's a fix that clarifies what is already the obvious/logical/safe implementation of the spec, but especially on the role names I can see someone could maybe argue otherwise.

I said that but Teodora pointed me to language in the spec that pretty clearly implies my reading is correct, and role names are meant to be unique:

> In order to accommodate prioritized delegations, the "roles" key in the DELEGATIONS object above points to an array of delegated roles, rather than to a hash table.

So delegations is an array only because "ordered hashtable" is not a thing in json.

joshuagl (approved):
Thanks for the change, this LGTM.

The `keyid`'s requirement is scoped only to the signatures field, and the `rolename` restriction does seem to gel with the intent of the rest of the specification.

@mnm678 WDYT?

mnm678 (approved):
This is a good clarification. In the future, I think we should discuss how to define uniqueness a bit more broadly (within both the delegating and delegated uses of keyids for example), but this change shouldn't block on that discussion.

tuf-spec.md
```diff
@@ -529,6 +529,8 @@ All signed metadata objects have the format:
       ::
         The identifier of the key signing the <a for="role">ROLE</a> object,
         which is a hexdigest of the SHA-256 hash of the canonical form of the key.
+        The keyid MUST be unique in the "signatures" dictionary: multiple
```
jku, line 532: Actually this should be `"signatures" array`

joshuagl (dismissed):

jku:
replaced "dictionary" with "array", sorry about doing that right after the reviews...

mnm678 (approved):

joshuagl (approved):

Issue #163, opened 2021-05-20
Add clarification on unknown key-value pairs in file format


jku:
Spec says:

> All of the formats described below include the ability to add more attribute-value fields for backwards-compatible format changes.

This does not make it very clear 
* that anyone de/serializing the format needs to be able to cope with undefined key-value pairs in various places in the file format: those key-value pairs are part of the data for the purposes of calculating hashes or signatures and MUST then be preserved if data is re-serialized
* where those key-value pairs can appear

My own understanding of the last point is that they can appear in every "object-like" dictionary at or below signed, meaning every dictionary in the format except ["hashes", "keys", "meta", "roles", "targets"].

asraa:
Nice! I have a use case for unknown key-values that @trishankatdatadog linked this issue to.

I completely agree that these key-value pairs must be preserved by an implementation so that hash/signature calculations can be consistent.

As far as where they can appear: my personal use case is key-values that may appear as auxiliary information in the KEYVAL dictionary (https://theupdateframework.github.io/specification/latest/#keyval). My linked PR adds a custom dictionary to this to append additional public key information (a certificate attesting to the public key).

trishankatdatadog:
Thanks, Asra! I think the next step is to add some text to the spec to clarify how to preserve and yet ignore these superset fields...

asraa:
Would it suffice to have it be similar to CUSTOM in targets (opaque to implementation, as a json object) and submit a PR in this repo with the change?

trishankatdatadog:
I don't think we even need to mention any specific field 🙂 We can add CUSTOM and your field as examples, but the important thing is to advise implementations what to do when seeing these added/superset fields. Makes sense?

PR #169, opened 2021-06-11
Add guidance about unknown fields

asraa:
https://github.com/theupdateframework/specification/issues/163

Addressing this, specify that implementers should preserve unknown key-value pairs and include them in calculating hashes or verifying signatures.

I would like to add examples here, but it feels like an awkward place if the format hasn't been defined at this point in the spec.

Signed-off-by: Asra Ali <asraa@google.com>

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-05-27
+Date: 2021-06-11
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.19
+Text Macro: VERSION 1.0.20
 </pre>
 
 Note: We strive to make the specification easy to implement, so if you come
@@ -490,9 +490,11 @@ A delegated role file is located at:
 # Document formats # {#document-formats}
 
 All of the formats described below include the ability to add more
-attribute-value fields for backwards-compatible format changes.  If
-a backwards incompatible format change is needed, a new filename can
-be used.
+attribute-value fields to dictionary objects for backwards-compatible format
+changes.  Implementers who encounter undefined key-value pairs in the format
+must include the data when calculating hashes or verifying signatures and must
+preserve the data when re-serializing. If a backwards incompatible format change
+is needed, a new filename can be used.
 
 ## Metaformat ## {#metaformat}
 
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -490,7 +490,7 @@ A delegated role file is located at:
 # Document formats # {#document-formats}
 
 All of the formats described below include the ability to add more
-attribute-value fields to dictionary objects for backwards-compatible format
+key-value fields to dictionary objects for backwards-compatible format
 changes.  Implementers who encounter undefined key-value pairs in the format
 must include the data when calculating hashes or verifying signatures and must
 preserve the data when re-serializing. If a backwards incompatible format change
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -490,8 +490,8 @@ A delegated role file is located at:
 # Document formats # {#document-formats}
 
 All of the formats described below include the ability to add more
-key-value fields to dictionary objects for backwards-compatible format
-changes.  Implementers who encounter undefined key-value pairs in the format
+attribute-value fields to objects for backwards-compatible format changes.
+Implementers who encounter undefined attribute-value pairs in the format
 must include the data when calculating hashes or verifying signatures and must
 preserve the data when re-serializing. If a backwards incompatible format change
 is needed, a new filename can be used.
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-06-11
+Date: 2021-06-14
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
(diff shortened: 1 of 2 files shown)


mnm678:
Thanks @asraa! The text of the change looks great, but could you remove the whitespace changes? They make the diff hard to read.

asraa:
Sorry about that! Must've been auto-formatting.

tuf-spec.md
```diff
@@ -490,9 +490,11 @@ A delegated role file is located at:
 # Document formats # {#document-formats}
 
 All of the formats described below include the ability to add more
-attribute-value fields for backwards-compatible format changes.  If
-a backwards incompatible format change is needed, a new filename can
-be used.
+attribute-value fields to dictionary objects for backwards-compatible format
```
trishankatdatadog, line 493:
```suggestion
key-value fields to dictionary objects for backwards-compatible format
```

tuf-spec.md
```diff
@@ -490,9 +490,11 @@ A delegated role file is located at:
...
+changes.  Implementers who encounter undefined key-value pairs in the format
+must include the data when calculating hashes or verifying signatures and must
+preserve the data when re-serializing. If a backwards incompatible format change
```
trishankatdatadog, line 496: I'm not sure what this last sentence means exactly, but we'll leave it for another PR, since you didn't write it. Maybe someone should make an issue to try to clarify it, though.

trishankatdatadog (commented):
I can't think of how this PR affects security, at least for TUF purposes. Did I miss anything? @mnm678 @SantiagoTorres @JustinCappos

tuf-spec.md
mnm678: This looks like a sentence left from pre-semantic versioning. I opened #170.

mnm678:
> I can't think of how this PR affects security, at least for TUF purposes. Did I miss anything?

Well, someone could try some kind of code injection through the unknown field, but it's signed metadata, so that's pretty unlikely.

asraa:
> Well, someone could try some kind of code injection through the unknown field, but it's signed metadata, so that's pretty unlikely.

This would be like someone adding an strange field that disappears when canonicalizing and verifying the signature, right? Did this ever have some discussion with regards to the CUSTOM targets dictionary? I suppose one mitigation is that users must consume canonicalized custom key-values or something like that.

trishankatdatadog:
> This would be like someone adding an strange field that disappears when canonicalizing and verifying the signature, right? Did this ever have some discussion with regards to the CUSTOM targets dictionary? I suppose one mitigation is that users must consume canonicalized custom key-values or something like that.

I think Marina means something else, but I'm not sure what

mnm678:
I was thinking of a json parsing vulnerability. But any of these attacks would require an attacker to tamper with signed metadata, so there shouldn't be a security implication.

trishankatdatadog:
> I was thinking of a json parsing vulnerability. But any of these attacks would require an attacker to tamper with signed metadata, so there shouldn't be a security implication.

Right, and any attacker who controls the key can do this without unknown key-value fields anyway.

Asra, what attack did you have in mind?

asraa:
> Asra, what attack did you have in mind?

I thought similar -- as in an attacker adds some specially crafted fields that when canonicalized disappear, so the signature is still verified, by a client thinks there's some extra "verified" data. This may not be possible though (and is obviously visible in the files)

trishankatdatadog:
> I thought similar -- as in an attacker adds some specially crafted fields that when canonicalized disappear, so the signature is still verified, by a client thinks there's some extra "verified" data. This may not be possible though (and is obviously visible in the files)

Yeap, the good news is that they will be caught, provided we log everything that goes in and out 🙂 

The safest thing to do is to explicitly disallow superset fields, but I'm afraid this is too restrictive for valid, real-world use cases.

I can see how TUF metadata can be used to carry other things that might lead to insecurity, but I don't see how that is directly a problem, so long as it does not break TUF's security guarantees. This is already doable with the targets themselves anyway, not to mention the custom targets metadata.

trishankatdatadog:
BTW, this has implications for the [sigstore key ceremony](https://github.com/sigstore/root-signing) this Friday, which uses superset fields to record YubiKey attestations alongside public keys in their TUF root metadata. So, if anyone is really concerned, please speak up ASAP.

tuf-spec.md
joshuagl: The rest of the spec is fairly consistent in referring to them as attributes and attribute-value fields.

joshuagl: I think for this PR we should revert to attribute-value fields, to be consistent.

We might want to edit the entire spec to change the object related terminology. Filed as https://github.com/theupdateframework/specification/issues/175

trishankatdatadog:
I can't think of major security issues right now with this change. Does anyone see issues? @JustinCappos?

joshuagl:
I can't think of a security issue with the change either, I did suggest we change some phrasing to be more consistent with the rest of the spec https://github.com/theupdateframework/specification/pull/169#discussion_r672284490 (but I also prefer the phrasing used in this PR, and think we should convert the entire spec to use it https://github.com/theupdateframework/specification/issues/175)

trishankatdatadog (dismissed):
Approved, but I want everyone here to think about this

tuf-spec.md
```diff
@@ -490,9 +490,11 @@ A delegated role file is located at:
 # Document formats # {#document-formats}
 
 All of the formats described below include the ability to add more
-attribute-value fields for backwards-compatible format changes.  If
-a backwards incompatible format change is needed, a new filename can
-be used.
+key-value fields to dictionary objects for backwards-compatible format
+changes.  Implementers who encounter undefined key-value pairs in the format
+must include the data when calculating hashes or verifying signatures and must
+preserve the data when re-serializing. If a backwards incompatible format change
```
joshuagl, line 496:
```suggestion
attribute-value fields to objects for backwards-compatible format changes.
Implementers who encounter undefined attribute-value pairs in the format
must include the data when calculating hashes or verifying signatures and must
preserve the data when re-serializing. If a backwards incompatible format change
```

The spec consistently uses attribute-value and object (rather than key-value and dictionary). Let's use the standard terms until https://github.com/theupdateframework/specification/issues/175 is addressed.

joshuagl (approved):
LGTM, thanks @asraa – just a minor suggestion to use the specification standard terms.

mnm678 (approved):

trishankatdatadog (approved):

trishankatdatadog (approved):

mnm678 (approved):

Issue #185, opened 2021-09-16
Delineating between security property requirements and internal consistency requirements in detailed client workflow


joshuagl:
There are parts of the specification which implementers SHOULD adhere to in order to gain the security properties of TUF.
There are parts of the specification which implementers need not strictly adhere to, so long as they are internally consistent across the detailed client workflow.

For example, it is not strictly necessary to persist metadata files in the recommended fixed form. However, it is necessary to store the data in a consistent and consistently identifiable fashion in order to be able to implement the workflow.

It could be helpful if the specification explicitly delineated between these two types of requirement so that implementers can more easily make informed decisions. In many cases the delineation is implicit (through use of RFC 2119 keywords on pieces of the detailed client workflow related to TUF security properties), but more explicit notation/annotations would make it easier for integrators to adopt TUF safely.

This may be easier to do after #121

mnm678:
> It could be helpful if the specification explicitly delineated between these two types of requirement so that implementers can more easily make informed decisions. In many cases the delineation is implicit (through use of RFC 2119 keywords on pieces of the detailed client workflow related to TUF security properties), but more explicit notation/annotations would make it easier for integrators to adopt TUF safely.

I think part of this effort should be to make more formal use of the RFC 2119 keywords (ie MUST/SHALL for security critical features, SHOULD for features that are recommended, MAY for suggestions). This approach as worked well for Uptane in helping define Uptane compliance.

PR #187, opened 2021-09-21
Clean up "Writing consistent snapshots" section

### LINKED ISSUE #184: should repository make unversioned root metadata available or not?
small inconsistency with regards to root.json versions: _6.2.1. Writing consistent snapshots_ states:

> an implementation must write both root.json and version_number.root.json because it is possible to download root metadata both with and without known version numbers.

It is _not_ in fact documented that client could download root without a version (client workflow only downloads versioned roots). I can see how a Trust-On-First-Use (TOFU) client initialization could use this but that is not defined in the spec.

I guess either
* TOFU initialization could be mentioned as a possibility (even if not recommended)
* the requirement to offer unversioned root.json should not be there

joshuagl:
* Match how metadata and target filenames are referred to in the detailed
  client workflow, i.e. filename.ext -> FILENAME.EXT
* Remove suggestion to write a version of the root metadata without a
  version prefix. We only ever look for version-prefixed version of root
  metadata.

Fixes #184

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-09-07
+Date: 2021-09-21
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, VMware
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.25
+Text Macro: VERSION 1.0.26
 </pre>
 
 Note: We strive to make the specification easy to implement, so if you come
@@ -1586,12 +1586,12 @@ We now explain how a repository should write metadata and targets to
 produce self-contained consistent snapshots.
 
 Simply put, TUF should write every metadata file as such: if the
-file had the original name of filename.ext, then it should be written to
-non-volatile storage as version_number.filename.ext, where version_number
-is an integer.
+file had the original name of FILENAME.EXT, then it should be written to
+non-volatile storage as VERSION_NUMBER.FILENAME.EXT, where VERSION_NUMBER
+is the integer version number listed in the metadata file.
 
 On the other hand, consistent target files should be written to
-non-volatile storage as digest.filename.ext.  This means that if the
+non-volatile storage as HASH.FILENAME.EXT.  This means that if the
 referrer metadata lists N cryptographic hashes of the referred file, then
 there must be N identical copies of the referred file, where each file will
 be distinguished only by the value of the digest in its filename. The
@@ -1600,16 +1600,11 @@ function used to produce the digest because, on a read, the choice of
 function follows from the selection of a digest (which includes the name of
 the cryptographic function) from all digests in the referred file.
 
-Additionally, the timestamp metadata (timestamp.json) should also be
-written to non-volatile storage whenever it is updated. It is OPTIONAL for
-an implementation to write identical copies at
-version_number.timestamp.json for record-keeping purposes, because a
-cryptographic hash of the timestamp metadata is usually not known in
-advance. The same step applies to the root metadata (root.json), although
-an implementation must write both root.json and version_number.root.json
-because it is possible to download root metadata both with and without
-known version numbers. These steps are required because these are the only
-metadata files that may be requested without known version numbers.
+Timestamp metadata (timestamp.EXT) SHOULD be written to non-volatile storage
+without a version prefix whenever it is updated. This is required because
+timestamp metadata is the only metadata file that may be requested without known
+version numbers.  It is OPTIONAL for an implementation to write identical copies
+at VERSION_NUMBER.timestamp.EXT for record-keeping purposes.
 
 Most importantly, no metadata file format must be updated to refer to the
 names of metadata or target files with their version numbers included. In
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1600,16 +1600,11 @@ function used to produce the digest because, on a read, the choice of
...
-written to non-volatile storage whenever it is updated. It is OPTIONAL for
-an implementation to write identical copies at
-version_number.timestamp.json for record-keeping purposes, because a
-cryptographic hash of the timestamp metadata is usually not known in
-advance. The same step applies to the root metadata (root.json), although
-an implementation must write both root.json and version_number.root.json
-because it is possible to download root metadata both with and without
-known version numbers. These steps are required because these are the only
-metadata files that may be requested without known version numbers.
+Timestamp metadata (timestamp.EXT) SHOULD be written to non-volatile storage
```
mnm678, line 1603: per RFC 2119, I think this should be a MUST

tuf-spec.md
```diff
@@ -1586,12 +1586,12 @@ We now explain how a repository should write metadata and targets to
...
-file had the original name of filename.ext, then it should be written to
-non-volatile storage as version_number.filename.ext, where version_number
-is an integer.
+file had the original name of FILENAME.EXT, then it should be written to
+non-volatile storage as VERSION_NUMBER.FILENAME.EXT, where VERSION_NUMBER
+is the integer version number listed in the metadata file.
 
 On the other hand, consistent target files should be written to
-non-volatile storage as digest.filename.ext.  This means that if the
+non-volatile storage as HASH.FILENAME.EXT.  This means that if the
```
mnm678, line 1594: In general I like `DIGEST` better than `HASH`, but I know that we use `HASH` elsewhere in the spec. No need to change it unless others feel strongly.

tuf-spec.md
```diff
@@ -1586,12 +1586,12 @@ We now explain how a repository should write metadata and targets to
 produce self-contained consistent snapshots.
 
 Simply put, TUF should write every metadata file as such: if the
```
mnm678, line 1588: I think this and the below paragraph should also be MUSTs.

tuf-spec.md
joshuagl: I will make that change. I was thinking of SHOULD in the context of #185, but I should handle that separately.

tuf-spec.md
joshuagl: I used `HASH` because that's what we use in the the [5.7.3](https://theupdateframework.github.io/specification/v1.0.25/#fetch-target) of the detailed client workflow and the name of the attribute in metadata is 'hashes'.

tuf-spec.md
joshuagl: I've added quite a few more RFC 2119 keywords in 6c54aa8 and cleaned up other paragraphs in the section.

joshuagl:
Thanks for the review @mnm678, I've addressed your comments and clarified some more of the text in the "Writing consistent snapshots" section. Please take a look.

mnm678 (approved):

joshuagl:
@lukpueh @trishankatdatadog could we get a second review here? Thank you!

tuf-spec.md
```diff
@@ -1585,49 +1585,44 @@ without interrupting that client.
...
-Finally, the root metadata should write the Boolean "consistent_snapshot"
-attribute at the root level of its keys of attributes. If consistent
-snapshots are not written by the repository, then the attribute may either
-be left unspecified or be set to the False value.  Otherwise, it must be
-set to the True value.
+Timestamp metadata (timestamp.EXT) MUST be written to non-volatile storage
+without a version prefix whenever it is updated. This is required because
+timestamp metadata is the only metadata file that may be requested without known
+version numbers.  It is OPTIONAL for an implementation to write identical copies
+at VERSION_NUMBER.timestamp.EXT for record-keeping purposes.
```
lukpueh, line 1607: Just took me a while to parse the sentence. For me it would have been helpful to read e.g. like so

```suggestion
version numbers.  It is OPTIONAL for an implementation to write identical copies of timestamp.EXT to the respective VERSION_NUMBER.timestamp.EXT for record-keeping purposes.
```

lukpueh (approved):
Great set of clarifications, @joshuagl! I added one minor suggestion. Feel free to ignore if you think the phrase is obvious without it.

joshuagl:
Thanks for the review both. I committed Lukas' change, which dismissed existing reviews. Please take another look @mnm678 @lukpueh

lukpueh (approved):

mnm678 (dismissed):

tuf-spec.md
```diff
@@ -1585,49 +1585,45 @@ without interrupting that client.
...
-
-Finally, the root metadata should write the Boolean "consistent_snapshot"
-attribute at the root level of its keys of attributes. If consistent
-snapshots are not written by the repository, then the attribute may either
-be left unspecified or be set to the False value.  Otherwise, it must be
-set to the True value.
+Timestamp metadata (timestamp.EXT) MUST be written to non-volatile storage
+without a version prefix whenever it is updated. This is required because
+timestamp metadata is the only metadata file that may be requested without known
+version numbers.  It is OPTIONAL for an implementation to write identical copies
```
trishankatdatadog, line 1606:
```suggestion
version numbers.  It is OPTIONAL for an implementation to write an identical copy
```

trishankatdatadog (approved):

tuf-spec.md
joshuagl: 💯 thanks

joshuagl:
Thanks for the suggestion Trishank. I've committed it and dismissed all reviews. Please take another look all @mnm678 @lukpueh @trishankatdatadog

lukpueh (approved):
Still LGTM!

mnm678 (approved):

PR #195, opened 2021-12-14
A round of clean-up/clarifications

joshuagl:
A few minor clean-up/clarification patches based on some recent reading and discussion of the spec:
1. consistent capitalisation of TAP
2. clarify that mirrors are optional
3. briefly describe the requirement for uniquely addressable targets in the repository layout section

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -399,7 +399,7 @@ publicly available in the [TAP directory](https://github.com/theupdateframework/
 
 An application uses the framework to interact with one or more repositories.
 A repository is a conceptual source of target files of interest to the
-application.  Each repository MAY have one or more mirrors which are the actual
+application.  Each repository MAY have one or more mirrors as the
 providers of files to be downloaded.  For example, each mirror may specify a
 different host where files can be downloaded from over HTTP.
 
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

tuf-spec.md
```diff
@@ -399,7 +399,7 @@ publicly available in the [TAP directory](https://github.com/theupdateframework/
 
 An application uses the framework to interact with one or more repositories.
 A repository is a conceptual source of target files of interest to the
-application.  Each repository has one or more mirrors which are the actual
+application.  Each repository MAY have one or more mirrors which are the actual
```
lukpueh, line 402: The phrase is a bit ambiguous. It could be read in way that target files are always only provided via mirrors, which is a bit confusing if mirrors are optional. Would it be clearer to say e.g.
```suggestion
application.  Each repository MAY have one or more mirrors as the
```
?

lukpueh (approved):
LGTM, I have one suggestions but it's not a blocker.

tuf-spec.md
joshuagl: This phrasing works for me, thanks @lukpueh

mnm678 (approved):

lukpueh (approved):

PR #196, opened 2021-12-14
Remove ambiguity in update root after a failed attempt

joshuagl:
In section [5.3. Update the root role](https://theupdateframework.github.io/specification/v1.0.26/#update-root) the specification suggests, in multiple steps (5.3.4, 5.3.5 and 5.3.10) that should the client fail to verify the updated root metadata the downloaded data should be discarded, the error reported, and
> On the next update cycle, begin at step § 5.3 Update the root role and version N of the root metadata file.

This directive only makes sense if the client application continues
running between update cycles and therefore still has initial trusted root
metadata loaded and a fixed update start time recorded.
For at least python-tuf and go-tuf this is not how the update workflow is
implementated.

Avoid confusion by removing the recommendation to start at 5.3 on the next
update cycle and instead leave only the suggestion to remove unverified
data. Logically, the next update cycle starts at the first step -- loading
the initial trusted root metadata.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-09-21
+Date: 2021-12-14
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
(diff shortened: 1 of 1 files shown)


mnm678 (approved):

lukpueh (approved):
LGTM. But should be merged after #195, or change the version number.

Issue #114, opened 2020-09-17
How should we handle metadata that has the same version but different content?


erickt:
The spec is a little vague as to how we should handle updating metadata that shares the same version number as the trusted metadata, but has different content. We can encounter this situation in two cases:
 
* When we fetch a new timestamp, section 5.2.2 says it's okay if the new file shares the same version as the trusted version. However, it doesn't say what we should do if this new file points at a different snapshot than the trusted timestamp.
* When we fetch a target or delegated target, section 4.4.4 states it's optional for the snapshot role to contain hashes of the targets and delegated targets. Without hashes, we have no way to tell without downloading the file to see if it matches the trusted version, nor what we should do if it doesn't match.

I can think of a few possible ways we could extend the spec to cover these situations:

* Just trust the new file (this though would probably expose us to all sorts of weird issues).
* Always try to download metadata, even if we think we have that version locally. Check that the new file matches our trusted content, or fail the update.
* Implement the above test for timestamp updates, but instead in section 5.4, change the logic to skip fetching any metadata version if we already trust that version (and hash if it's listed in the trusted snapshot).

rdimitrov:
@erickt - Hey, I think #209 partially covered the topics of this issue, at least in regards to the timestamp update process. Are the rest of the thoughts still relevant or perhaps we can close this one?

PR #209, opened 2022-02-16
Update metadata version comparison rules in client workflow

### LINKED ISSUE #207: Verify and fix the version comparison done for metadata files
**Description:**

- Verify the usage of “less than vs less than or equal to” done for version comparison throughout the specification
- The confusion comes that the spec uses the same text in multiple scenarios that are not the same
- Meta versions can be equal, but metadata versions probably should not be. For everything other than timestamp, it does not matter. For timestamp, spec always expects to download the latest and use that
- For example, it’s normal for the timestamp version to change but still point to the same snapshot version

**Related issues:**

- https://github.com/theupdateframework/go-tuf/issues/218

rdimitrov:
The client workflow has a set of version comparisons rules for how
to update metadata files. The following PR addresses the differences
coming from the fact that when updating not all metadata files should
be treated equally.

Fixes #207 and is related to #114 

Notes:
- **root** should be updated only when there's a newer version
- **timestamp** should be updated only when there's a newer version AND the referenced **snapshot** version is not less than the one in the already trusted timestamp
- **snapshot** should be updated only when its version is the same as the one referenced in the trusted timestamp AND the version of the referenced **targets** metadata (incl. delegated targets files) is equal or newer than the ones in the trusted snapshot.
- **targets** should be updated only when its version is the same as the one referenced in the trusted **snapshot**
- In regards to #114, for timestamp, the decision is to take into account the new timestamp metadata file **only** if its version is not less than or equal to the trusted one, so this PR addresses one of the use cases mentioned in the issue.

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1315,11 +1315,11 @@ it in the next step.
   and report the signature failure.
 
 5. **Check for a rollback attack.** The version number of the trusted
-  root metadata file (version N) MUST be less than or equal to the version
+  root metadata file (version N) MUST be less than the version
   number of the new root metadata file (version N+1). Effectively, this means
   checking that the version number signed in the new root metadata file is
-  indeed N+1.  If the version of the new root metadata file is less than the
-  trusted metadata file, discard it, abort the update cycle, and report the
+  indeed N+1.  If the version of the new root metadata file is less than or equal
+  to the version of the trusted metadata file, discard it, abort the update cycle, and report the
   rollback attack.
 
 6. Note that the expiration of the new (intermediate) root metadata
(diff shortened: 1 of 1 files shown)


mnm678 (dismissed):

tuf-spec.md
```diff
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2021-12-14
+Date: 2022-02-17
```
joshuagl, line 6:
```suggestion
Date: 2022-04-28
```

tuf-spec.md
```diff
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.28
+Text Macro: VERSION 1.0.29
```
joshuagl, line 19:
```suggestion
Text Macro: VERSION 1.0.30
```

joshuagl (approved):
This has long been a source of confusion in the spec. Thank you for the fix!

joshuagl (approved):

joshuagl:
@mnm678 if you're still comfortable with the change could you re-approve? Thank you.

mnm678 (approved):

PR #224, opened 2022-05-09
feat: add reusable workflow for autofilling an issue when there's a new TUF version

rdimitrov:
The following adds a workflow that can be used by projects that want to keep track and be notified when there's a new version of the TUF specification. 

In case there's a newer version, it will file an issue against the project with several links, like the currently supported version, and a comparison showing what's changed. Here's an example of the issue it will create - 

**Title:** 
> TUF specification has a new version - v1.0.30

**Body:** 
> Hey, it seems there's a newer version of the TUF specification - [v1.0.30](https://github.com/theupdateframework/specification/blob/v1.0.30/tuf-spec.md)
> 
> The version which [rdimitrov/python-tuf](https://github.com/rdimitrov/python-tuf) state it supports is - [v1.0.29](https://github.com/theupdateframework/specification/blob/v1.0.29/tuf-spec.md)
> 
> The following is a comparison of what changed between the two versions - [Compare v1.0.29 to v1.0.30](https://github.com/theupdateframework/specification/compare/v1.0.29...v1.0.30)
> 
> Please review the newer version and address the changes.

**Note:** 
A few other PRs will be created and referenced here for https://github.com/theupdateframework/python-tuf and https://github.com/theupdateframework/go-tuf which can serve as examples of how to use this workflow.

Related to - https://github.com/theupdateframework/go-tuf/issues/283

### DIFF
--- a/.github/workflows/check-latest-spec-version.yml
+++ b/.github/workflows/check-latest-spec-version.yml
@@ -19,12 +19,12 @@ jobs:
               owner: "theupdateframework",
               repo: "specification"
             });
-            const spec_version = release.data.tag_name
+            const latest_version = release.data.tag_name
             const supported_version = "${{ inputs.tuf-version }}"
             const repo = context.repo.owner + "/" + context.repo.repo
-            if (spec_version != supported_version) {
+            if (latest_version != supported_version) {
               console.log(
-                "The latest TUF specification version (" + spec_version + ") does not match with the version that " + repo + " supports (" +
+                "The latest TUF specification version (" + latest_version + ") does not match with the version that " + repo + " supports (" +
                   supported_version + ")"
               )
               const issues = await github.rest.search.issuesAndPullRequests({
@@ -36,12 +36,14 @@ jobs:
                 await github.rest.issues.create({
                   owner: context.repo.owner,
                   repo: context.repo.repo,
-                  title: "TUF specification has a new version - " + spec_version,
-                  body: "Hey, it seems there's a newer version of the TUF specification - [" + spec_version +
-                        "](https://github.com/theupdateframework/specification/blob/" + spec_version + "/tuf-spec.md)" +
+                  title: "TUF specification has a new version - " + latest_version,
+                  body: "Hey, it seems there's a newer version of the TUF specification - [" + latest_version +
+                        "](https://github.com/theupdateframework/specification/blob/" + latest_version + "/tuf-spec.md)" +
                         "\n\n" + 
-                        "For comparison, the version which [" + repo + "](https://github.com/" + repo + ") state it supports is - [" + supported_version +
+                        "The version which [" + repo + "](https://github.com/" + repo + ") state it supports is - [" + supported_version +
                         "](https://github.com/theupdateframework/specification/blob/" + supported_version + "/tuf-spec.md)" + 
+                        "\n\n" +
+                        "The following is a comparison of what changed between the two versions - [Compare " + supported_version + " to " + latest_version +"](https://github.com/theupdateframework/specification/compare/" + supported_version + "..." + latest_version + ")" +
                         "\n\n" +
                         "Please review the newer version and address the changes."
                 })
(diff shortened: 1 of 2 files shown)


jku:
I have a question on permissions: can the python-tuf workflow define a permission so that the embedded action can't override that?

I'm asking in context of the `@master` versioning in the workflow... If python-tuf can't practically limit the permission, then a compromised specification repo would immediately compromise the python-tuf repo as well if we use `@master`

rdimitrov:
> I have a question on permissions: can the python-tuf workflow define a permission so that the embedded action can't override that?
> 
> I'm asking in context of the `@master` versioning in the workflow... If python-tuf can't practically limit the permission, then a compromised specification repo would immediately compromise the python-tuf repo as well if we use `@master`
> 
> link: [theupdateframework/python-tuf#2001 (comment)](https://github.com/theupdateframework/python-tuf/pull/2001#discussion_r868881109)

Yes, you should be able to either set the permissions for a specific job or for the whole workflow 👍 

ref: [Github Actions/permissions](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)

joshuagl (commented):
Can we add a comment to the top of the reusable workflow which includes a one-line description of what the workflow does and includes an example of how to use it?

It might also be worth mentioning in the README that the reusable workflow exists and implementers can use it to be notified of when new specification versions are released? This could also be a follow-on PR if preferred.

rdimitrov:
> Can we add a comment to the top of the reusable workflow which includes a one-line description of what the workflow does and includes an example of how to use it?
> 
> It might also be worth mentioning in the README that the reusable workflow exists and implementers can use it to be notified of when new specification versions are released? This could also be a follow-on PR if preferred.

Thanks! I've addressed the suggestions 👍

joshuagl (approved):
This is great, thanks!

joshuagl:
Could we get a second review here from an @theupdateframework/tap-editors ?

JustinCappos:
This is a great idea!  I'd love to see maintainers of other TUF
implementations comment as well, but I am very supportive overall.

lukpueh (approved):
This is really great! And it LGTM, modulo a copy-pasted typo, some trailing whitespace and a line-wrap nit. Feel free to address or ignore. I'll approve as is.

Note, I also diffed the script with python-tuf's [`specification-version.yml`](https://github.com/theupdateframework/python-tuf/blob/211f2afe56c61e0eac35d338081a532bd12dc4ed/.github/workflows/specification-version.yml), which we know works and probably served as a base for this workflow(?). It looks like they do the same thing.

rdimitrov:
> This is really great! And it LGTM, modulo a copy-pasted typo, some trailing whitespace and a line-wrap nit. Feel free to address or ignore. I'll approve as is.
> 
> Note, I also diffed the script with python-tuf's [`specification-version.yml`](https://github.com/theupdateframework/python-tuf/blob/211f2afe56c61e0eac35d338081a532bd12dc4ed/.github/workflows/specification-version.yml), which we know works and probably served as a base for this workflow(?). It looks like they do the same thing.

Thanks for the review! 👍 I've used it as a base so thus the reason for the PR in python-tuf which discards the old one and switches to the one that will be hosted here - https://github.com/theupdateframework/python-tuf/pull/2001

joshuagl (approved):

lukpueh:
> > This is really great! And it LGTM, modulo a copy-pasted typo, some trailing whitespace and a line-wrap nit. Feel free to address or ignore. I'll approve as is.
> > Note, I also diffed the script with python-tuf's [`specification-version.yml`](https://github.com/theupdateframework/python-tuf/blob/211f2afe56c61e0eac35d338081a532bd12dc4ed/.github/workflows/specification-version.yml), which we know works and probably served as a base for this workflow(?). It looks like they do the same thing.
> 
> Thanks for the review! 👍 I've used it as a base so thus the reason for the PR in python-tuf which discards the old one and switches to the one that will be hosted here - [theupdateframework/python-tuf#2001](https://github.com/theupdateframework/python-tuf/pull/2001)

Oh yeah, it was not meant as a criticism, but more as a *this has already been reviewed elsewhere, so it should be fine* statement. :)

Issue #69, opened 2019-12-03
How should clients handle interrupted updates?


erickt:
I am currently doing some exploration into how clients should handle interrupted, partially successful updates. For example, say we have a client that has a local cached copy of valid and  unexpired metadata.  We start an update process which includes a new `timestamp`, `snapshot`, and `targets` metadata. Unfortunately, we download the new timestamp and snapshot and persist them to disk, but the device loses power. Then when power is restored the network is down. We'd still like to make queries against the TUF targets file, but according to the workflow, we should get an error. We can only recover from this by restoring the network.

This is particularly relevant to Fuchsia, because of how we have created our packaging system. We want to treat the TUF targets as the list of executable packages, since it allows us to maintain a cryptographic chain of trust all the way down to the bootloader for what things can be executed. All our packages are stored in a content-addressed filesystem, and we use the `custom` field in a TUF target to provide the mapping from a human readable name to a merkle addressed package blob. When we try to open a package, we first look in TUF to find the merkle, then we check if we've already downloaded that blob. If so, we open up that package and serve it to the caller. See this [slightly stale doc](https://fuchsia.dev/fuchsia-src/development/workflows/package_update) for more details. Due to this interrupted update problem, there's a chance a Fuchsia device could be made unusable until we are able to finish updating our metadata. 

If not, we have had a few ideas on how to approach this:

* If an update fails, we could still query the local latest `targets` metadata, assuming it was signed with a key that's still trusted by the `root` metadata.
* During the update, we delay writing all the metadata to disk until all the files have been downloaded and verified. Then the files are written in one atomic transaction.
* For consistent snapshot metadata (which we only plan on supporting), fetch the `timestamp` metadata, but don't persist it to disk yet. Fetch and write the versioned-prefixed `snapshot` and `targets` metadata, and any other delegated metadata, to disk. Atomically write the `timestamp` metadata to disk, then clean up any old snapshot/targets/etc metadata files.

I'm not sure if these ideas would weaken the TUF security model though. Is there a better way for dealing with this, and could we incorporate this into the spec (or a POUF?), since I imagine other folks might need a solution for this.

joshuagl:
I think option 2 above makes more sense:
> During the update, we delay writing all the metadata to disk until all the files have been downloaded and verified. Then the files are written in one atomic transaction.

[...]

PR #226, opened 2022-05-19
Check timestamp/snapshot contains snapshot/targets version

erickt:
The [timestamp.json] states that the timestamp `meta` section MUST only contain a description of the snapshot.json file. However, [updating the timestamp role] does not describe when to perform this verification.

Similarly, the [snapshot.json] states that the snapshot `meta` section MUST contain a description of the targets.json file, which is also not described in [updating the snapshot role].

This patch explicitly states that these checks should be performed, and that the metadata should be rejected if it is missing these entries.

[timestamp.json]: https://theupdateframework.github.io/specification/v1.0.30/#file-formats-timestamp
[updating the timestamp role]: https://theupdateframework.github.io/specification/v1.0.30/#update-timestamp
[snapshot.json]: https://theupdateframework.github.io/specification/v1.0.30/#file-formats-snapshot
[updating the snapshot role]: https://theupdateframework.github.io/specification/v1.0.30/#update-snapshot

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -1368,20 +1368,24 @@ it in the next step.
 
 3. **Check for a rollback attack.**
 
-  1. The version number of the trusted timestamp metadata file, if
-    any, MUST be less than the version number of the new timestamp
-    metadata file. If the new timestamp metadata version is less than the trusted
-    timestamp metadata version, discard it, abort the update cycle, and
-    report the potential rollback attack. In case they are equal, discard the new
-    timestamp metadata and abort the update cycle. This is normal and it
-    shouldn't raise any error. The reason for aborting the update process is that
-    there shouldn't be any changes in the content of this, or any other metadata
-    files too, considering it has the same version as the already trusted one.
+  1. The version number of the trusted timestamp metadata file, if any, MUST be
+     less than the version number of the new timestamp metadata file. If the
+     new timestamp metadata version is less than the trusted timestamp metadata
+     version, discard it, abort the update cycle, and report the potential
+     rollback attack. In case they are equal, discard the new timestamp
+     metadata and abort the update cycle. This is normal and it shouldn't raise
+     any error. The reason for aborting the update process is that there
+     shouldn't be any changes in the content of this, or any other metadata
+     files too, considering it has the same version as the already trusted one.
 
-  2. The version number of the snapshot metadata file in the
-    trusted timestamp metadata file, if any, MUST be less than or equal to its
-    version number in the new timestamp metadata file. If not, discard the new
-    timestamp metadata file, abort the update cycle, and report the failure.
+  2. The new timestamp metadata file MUST only contain the description of the
+     snapshot metadata file.  If not, discard the new snapshot metadata file,
+     abort the cycle, and report the failure.
+
+  3. The version number of the snapshot metadata file in the trusted timestamp
+     metadata file, if any, MUST be less than or equal to its version number in
+     the new timestamp metadata file.  If not, discard the new timestamp
+     metadata file, abort the update cycle, and report the failure.
 
 4. **Check for a freeze attack.** The expiration timestamp in the
   new timestamp metadata file MUST be higher than the fixed update start time.
@@ -1425,14 +1429,20 @@ it in the next step.
   in the trusted timestamp metadata.  If the versions do not match, discard the
   new snapshot metadata, abort the update cycle, and report the failure.
 
-5. **Check for a rollback attack**. The version number of the targets
-  metadata file, and all delegated targets metadata files, if any, in the
-  trusted snapshot metadata file, if any, MUST be less than or equal to its
-  version number in the new snapshot metadata file. Furthermore, any targets
-  metadata filename that was listed in the trusted snapshot metadata file, if
-  any, MUST continue to be listed in the new snapshot metadata file.  If any of
-  these conditions are not met, discard the new snapshot metadata file, abort
-  the update cycle, and report the failure.
+5. **Check for a rollback attack**.
+
+  1. The new snapshot metadata file MUST contain the description of the targets
+     metadata file.  If not, discard the new snapshot metadata file, abort the
+     cycle, and report the failure.
+
+  2. The version number of the targets metadata file, and all delegated targets
+     metadata files, if any, in the trusted snapshot metadata file, if any,
+     MUST be less than or equal to its version number in the new snapshot
+     metadata file. Furthermore, any targets metadata filename that was listed
+     in the trusted snapshot metadata file, if any, MUST continue to be listed
+     in the new snapshot metadata file.  If any of these conditions are not
+     met, discard the new snapshot metadata file, abort the update cycle, and
+     report the failure.
 
 6. **Check for a freeze attack**. The expiration timestamp in the
   new snapshot metadata file MUST be higher than the fixed update start time.


tuf-spec.md
```diff
@@ -1368,20 +1368,24 @@ it in the next step.
...
 
-  1. The version number of the trusted timestamp metadata file, if
-    any, MUST be less than the version number of the new timestamp
-    metadata file. If the new timestamp metadata version is less than the trusted
-    timestamp metadata version, discard it, abort the update cycle, and
-    report the potential rollback attack. In case they are equal, discard the new
-    timestamp metadata and abort the update cycle. This is normal and it
-    shouldn't raise any error. The reason for aborting the update process is that
-    there shouldn't be any changes in the content of this, or any other metadata
-    files too, considering it has the same version as the already trusted one.
```
JustinCappos, line 1379: I believe the only change here is line wrap / formatting.  Can you omit this change?

tuf-spec.md
```diff
@@ -1368,20 +1368,24 @@ it in the next step.
...
+     metadata and abort the update cycle. This is normal and it shouldn't raise
+     any error. The reason for aborting the update process is that there
+     shouldn't be any changes in the content of this, or any other metadata
+     files too, considering it has the same version as the already trusted one.
+
+  2. The new timestamp metadata file MUST only contain the description of the
+     snapshot metadata file.  If not, discard the new snapshot metadata file,
+     abort the cycle, and report the failure.
+
+  3. The version number of the snapshot metadata file in the trusted timestamp
```
JustinCappos, line 1385: Same here.  Please don't redo the line breaks as part of another change.

tuf-spec.md
```diff
@@ -1425,14 +1429,20 @@ it in the next step.
...
-  trusted snapshot metadata file, if any, MUST be less than or equal to its
-  version number in the new snapshot metadata file. Furthermore, any targets
-  metadata filename that was listed in the trusted snapshot metadata file, if
-  any, MUST continue to be listed in the new snapshot metadata file.  If any of
-  these conditions are not met, discard the new snapshot metadata file, abort
-  the update cycle, and report the failure.
+5. **Check for a rollback attack**.
+
+  1. The new snapshot metadata file MUST contain the description of the targets
+     metadata file.  If not, discard the new snapshot metadata file, abort the
```
JustinCappos, line 1435:
```suggestion
     metadata files (the top-level targets.EXT and all delegated roles.)  If not, discard the 
     new snapshot metadata file, abort the
```
Clarifying that this includes all targets files.

tuf-spec.md
```diff
@@ -1425,14 +1429,20 @@ it in the next step.
...
-  metadata file, and all delegated targets metadata files, if any, in the
-  trusted snapshot metadata file, if any, MUST be less than or equal to its
-  version number in the new snapshot metadata file. Furthermore, any targets
-  metadata filename that was listed in the trusted snapshot metadata file, if
-  any, MUST continue to be listed in the new snapshot metadata file.  If any of
-  these conditions are not met, discard the new snapshot metadata file, abort
-  the update cycle, and report the failure.
+5. **Check for a rollback attack**.
+
+  1. The new snapshot metadata file MUST contain the description of the targets
```
JustinCappos, line 1434:
```suggestion
  1. The new snapshot metadata file MUST contain the version numbers of the targets
```
The listed item is the version numbers, not a description.

JustinCappos (changes requested):
Mostly good.  Needs a few minor tweaks.

tuf-spec.md
```diff
@@ -1368,20 +1368,24 @@ it in the next step.
...
+     version, discard it, abort the update cycle, and report the potential
+     rollback attack. In case they are equal, discard the new timestamp
+     metadata and abort the update cycle. This is normal and it shouldn't raise
+     any error. The reason for aborting the update process is that there
+     shouldn't be any changes in the content of this, or any other metadata
+     files too, considering it has the same version as the already trusted one.
+
+  2. The new timestamp metadata file MUST only contain the description of the
+     snapshot metadata file.  If not, discard the new snapshot metadata file,
+     abort the cycle, and report the failure.
```
joshuagl, line 1383:
```suggestion
  2. The new timestamp metadata file MUST only contain the description of the
     snapshot metadata file.  If not, discard the new timestamp metadata file,
     abort the cycle, and report the failure.
```

Just to be clear, here we're checking that timestamp's `meta` lists _only_ snapshot?

tuf-spec.md
joshuagl: I assume that "description" was used as a short-hand for the version plus optional length and hashes (fields in a `METAFILES` object)? As this terminology is not used elsewhere we should certainly be more exploit here.

joshuagl (changes requested):

tuf-spec.md
joshuagl: Perhaps we could just say
>The new snapshot metadata file's `METAFILES` object MUST list all targets metadata (top-level targets.EXT all all delegated roles) including their version numbers, optionally their length and optionally their hashes. If not, discard the new snapshot metadata file, abort the 

etc.

tuf-spec.md
erickt: done

tuf-spec.md
erickt: done

erickt (commented):
I did another pass to add explicit instructions on how to handle validating the version, and optional length and optional hashes. How does it read?

erickt:
I just read through README.rst, and it looks like I was supposed to submit this against the `draft` branch. However that branch hasn't been touched since 2019. Should I change this to merge into that branch?

tuf-spec.md
```diff
@@ -1368,20 +1368,40 @@ it in the next step.
...
+    discard the new timestamp metadata file, abort the update cycle, and report
+    the failure.
+
+  4. If the new timestamp metadata file's [=metapath/VERSION=] number of the
+    snapshot metadata file is equal to the [=metapath/VERSION=] numbers in the
+    trusted snapshot metadata file:
+
+    1. The [=metapath/LENGTH=] in the new timestamp metadata file, if any, MUST
+      be equal to the [=metapath/LENGTH=] in the trusted timestamp file, if any.
+      If not, discard the new timestamp metadata file, abort the cycle, and
```
JustinCappos, line 1398: Why do we discard the new timestamp metadata?  Could an attacker later roll back the timestamp file?

tuf-spec.md
```diff
@@ -1368,20 +1368,40 @@ it in the next step.
...
+      report the failure.
+
+    2. For each entry in the new timestamp metadata file's [=metapath/HASHES=]
+      dictionary, if the key is present in the trusted timestamp metadata file,
+      the values MUST be equal.  If not, discard the new timestamp metadata
```
JustinCappos, line 1403: Same comment as above here.

tuf-spec.md
```diff
@@ -1425,14 +1445,34 @@ it in the next step.
...
-  metadata filename that was listed in the trusted snapshot metadata file, if
-  any, MUST continue to be listed in the new snapshot metadata file.  If any of
-  these conditions are not met, discard the new snapshot metadata file, abort
-  the update cycle, and report the failure.
+5. **Check for a rollback attack**.
+
+  1. The new snapshot metadata file's [=METAFILES=] object MUST contain a
+    [=snapshot/METAPATH=] entry for the targets metadata file, and all delegated targets
+    metadata files, if any, in the trusted snapshot metadata file.  If not,
+    discard the new snapshot metadata file, abort the cycle, and report the
```
JustinCappos, line 1453: Do we need to discard the snapshot file in this case?  Why?

JustinCappos (commented):
I'd like to understand better why we would discard the new snapshot / timestamp file if there is a mismatch.

tuf-spec.md
erickt: I'm not sure if it's necessarily resisting an attack, but more about enabling the optimization where we can skip downloading metadata if we already have it locally without putting our local trusted metadata in an inconsistent state (such as in #114 and #227).

It might be helpful if we work with an example. Say we have locally trusted metadata, with Timestamp T1 and Snapshot S1, which is at version N. The remote repository is at Timestamp T1 and Snapshot S2, which also is at version N.

When we perform a client update, we will:

* Download T2
* Verify it was signed correctly
* Verify that T2's snapshot description's version >= the trusted snapshot S1's version

At this point we could decide that we we already have the trusted Snapshot S1, so we could abort the update workflow. However this would be incorrect, because it's possible that Snapshot S2 could have different content than Snapshot S1. If Timestamp T2 contains the optional length and/or hashes of Snapshot S2. This means that if we create a brand new client based off the local trusted metadata, we cannot use the length/hashes from Timestamp T2 to verify Snapshot S1. This would then force us to throw away the Snapshot S1 and download the new S2 version.

This patch tries to avoid this issue by enforcing that we cannot get into this state by treating Timestamp T2 as invalid, since it doesn't match what we already trust.

This would be especially handy if we implement transactional downloading of metadata in #69 to make sure the local metadata doesn't get into an incomplete state.

tuf-spec.md
erickt: This was covered in the above comment, so lets track it there.

tuf-spec.md
erickt: This was covered in the above comment, so lets track it there.

lukpueh:
> I just read through README.rst, and it looks like I was supposed to submit this against the draft branch. 

The README says:  
> * For patch-type changes, pull requests may be submitted directly against the 'master' branch."

So this should be fine.

lukpueh:
> However that branch hasn't been touched since 2019.

See #228 for an update and #229 for a request to prevent a stale `draft` branch in the future.

tuf-spec.md
lukpueh: > The remote repository is at Timestamp T1 and Snapshot S2, which also is at version N.

Did you mean **T2**? You do say the client downloads **T2** below. I assume this is a typo here?

Just to make sure, do I understand correctly that in your example:
- T1 describes S1
- T2 describes S2
- T2 has a higher version than T1
- S2 has the same version as S1 
- S2 and S1 have different contents

This implies that the repo has *incorrectly* updated snapshot and timestamp, without bumping snapshot's version number, right?

Following you proposal, the client that already trusts T1 and S1 can only update, if the repo issues a new T2 (or higher) describing a new S2, with a version number higher than S1. Is that correct? If so, I think it's fair.

Issue #242, opened 2022-08-12
It would be nice to have an OpenAPI or JSON Schema for TUF resource types


patflynn:
While building Sigstore TUF clients it's necessary to parse the various resource types (Root, Snapshot, Target,..).  Most languages have serialization tools that can do this automatically with the right OpenAPI or Json Schema docs.  Making these available could accelerate client development.

PR #246, opened 2022-08-25
Add TUF and in-toto schema files

fridex:
Add TUF and in-toto JSON schema files. These schema files were produced as part of Datadog agent integration testsuite (additionally adjusted to remove Datadog specific parts) and have not been reviewed yet - I'm new to TUF/in-toto, so please review any possible inconsistencies and sorry for them.

Closes: #242

### DIFF
--- a/schemas/in-toto-metadata-signer-x.schema.json
+++ b/schemas/in-toto-metadata-signer-x.schema.json
@@ -1,0 +1,145 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "signatures": {
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "keyid": {
+              "type": "string",
+              "minLength": 64,
+              "maxLength": 64,
+              "pattern": "^[a-f0-9]{64}$"
+            },
+            "sig": {
+              "type": "string",
+              "minLength": 1024,
+              "maxLength": 1024,
+              "pattern": "^[a-f0-9]{1024}$"
+            }
+          },
+          "required": [
+            "keyid",
+            "sig"
+          ]
+        }
+      ]
+    },
+    "signed": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "_type": {
+          "enum": ["targets"]
+        },
+        "delegations": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "keys": {
+              "type": "object"
+            },
+            "roles": {
+              "type": "array",
+              "items": {}
+            }
+          },
+          "required": [
+            "keys",
+            "roles"
+          ]
+        },
+        "expires": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
+        },
+        "targets": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^(simple/[a-z-0-9]+/[a-z0-9-]+-py2\\.py3-none-any\\.whl)|(in-toto-metadata/[a-f0-9]{64}/((wheels-signer)|(tag)|(wheels-builder))\\.[a-f0-9]{8}\\.link)$": {
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "custom": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "in-toto": {
+                      "type": "array",
+                      "items": [
+                        {
+                          "type": "string",
+                          "pattern": "in-toto-metadata/([a-f0-9]{64}/(wheels-signer|wheels-builder|tag)\\.[a-f0-9]{8}\\.link)|root.layout"
+                        }
+                      ]
+                    },
+                    "root-layout-type": {
+                      "enum": ["extras", "core"]
+                    }
+                  }
+                },
+                "hashes": {
+                  "type": "object",
+                  "additionalProperties": false,
+                  "properties": {
+                    "sha256": {
+                      "type": "string",
+                      "minLength": 64,
+                      "maxLength": 64,
+                      "pattern": "^[a-f0-9]{64}$"
+                    },
+                    "sha512": {
+                      "type": "string",
+                      "minLength": 128,
+                      "maxLength": 128,
+                      "pattern": "^[a-f0-9]{128}$"
+                    }
+                  },
+                  "required": [
+                    "sha256",
+                    "sha512"
+                  ]
+                },
+                "length": {
+                  "type": "integer",
+                  "minimum": 1
+                }
+              },
+              "required": [
+                "custom",
+                "hashes",
+                "length"
+              ]
+            }
+          }
+        },
+        "version": {
+          "type": "integer",
+          "minimum": 0
+        }
+      },
+      "required": [
+        "_type",
+        "delegations",
+        "expires",
+        "spec_version",
+        "targets",
+        "version"
+      ]
+    }
+  },
+  "required": [
+    "signatures",
+    "signed"
+  ]
+}
--- a/schemas/in-toto-metadata-signer.schema.json
+++ b/schemas/in-toto-metadata-signer.schema.json
@@ -1,0 +1,168 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
+    "signatures": {
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "properties": {
+            "keyid": {
+              "type": "string",
+              "minLength": 64,
+              "maxLength": 64,
+              "pattern": "^[a-f0-9]{64}$"
+            },
+            "sig": {
+              "type": "string",
+              "minLength": 1024,
+              "maxLength": 1024,
+              "pattern": "^[a-f0-9]{1024}$"
+            }
+          },
+          "required": [
+            "keyid",
+            "sig"
+          ]
+        }
+      ]
+    },
+    "signed": {
+      "type": "object",
+      "properties": {
+        "_type": {
+          "enum": ["targets"]
+        },
+        "delegations": {
+          "type": "object",
+          "properties": {
+            "keys": {
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {
+                "^[a-f0-9]{64}$": {
+                  "type": "object",
+                  "properties": {
+                    "keyid_hash_algorithms": {
+                      "type": "array",
+                      "items": {
+                        "enum": [
+                          "sha256",
+                          "sha512"
+                        ]
+                      }
+                    },
+                    "keytype": {
+                      "enum": ["rsa"]
+                    },
+                    "keyval": {
+                      "type": "object",
+                      "properties": {
+                        "public": {
+                          "type": "string",
+                          "pattern": "^-----BEGIN PUBLIC KEY-----"
+                        }
+                      },
+                      "required": [
+                        "public"
+                      ]
+                    },
+                    "scheme": {
+                      "enum": ["rsassa-pss-sha256"]
+                    }
+                  },
+                  "required": [
+                    "keyid_hash_algorithms",
+                    "keytype",
+                    "keyval",
+                    "scheme"
+                  ]
+                }
+              }
+            },
+            "roles": {
+              "type": "array",
+              "items": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "keyids": {
+                      "type": "array",
+                      "items": [
+                        {
+                          "type": "string",
+                          "minLength": 64,
+                          "maxLength": 64,
+                          "pattern": "^[a-f0-9]{64}$"
+                        }
+                      ]
+                    },
+                    "name": {
+                      "type": "string",
+                      "pattern": "^in-toto-metadata-signer-[0-9a-f]$"
+                    },
+                    "paths": {
+                      "type": "array",
+                      "items": [
+                        {
+                          "type": "string",
+                          "pattern": "^in-toto-metadata/[0-9a-f]\\*/\\*\\.link$"
+                        }
+                      ]
+                    },
+                    "terminating": {
+                      "type": "boolean"
+                    },
+                    "threshold": {
+                      "type": "integer",
+                      "minimum": 1
+                    }
+                  },
+                  "required": [
+                    "keyids",
+                    "name",
+                    "paths",
+                    "terminating",
+                    "threshold"
+                  ]
+                }
+              ]
+            }
+          },
+          "required": [
+            "keys",
+            "roles"
+          ]
+        },
+        "expires": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$"
+        },
+        "targets": {
+          "type": "object"
+        },
+        "version": {
+          "type": "integer",
+          "minimum": 0
+        }
+      },
+      "required": [
+        "_type",
+        "delegations",
+        "expires",
+        "spec_version",
+        "targets",
+        "version"
+      ]
+    }
+  },
+  "required": [
+    "signatures",
+    "signed"
+  ]
+}
--- a/schemas/root.schema.json
+++ b/schemas/root.schema.json
@@ -1,0 +1,181 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "required": ["signatures", "signed"],
+  "additionalProperties": false,
+  "properties": {
+    "signatures": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["keyid", "sig"],
+        "additionalProperties": false,
+        "properties": {
+          "keyid": {
+            "type": "string",
+            "minLength": 64,
+            "maxLength": 64,
+            "pattern": "^[a-f0-9]{64}$"
+          },
+          "sig": {
+            "type": "string",
+            "minLength": 1024,
+            "maxLength": 1024,
+            "pattern": "^[a-f0-9]{1024}$"
+          }
+        }
+      }
+    },
+    "signed": {
+      "type": "object",
+      "required": ["_type", "consistent_snapshot", "expires", "keys", "roles", "spec_version", "version"],
+      "additionalProperties": false,
+      "properties": {
+        "_type": {
+          "enum": ["root"]
+        },
+        "consistent_snapshot": {
+          "type": "boolean"
+        },
+        "expires": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "keys": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[0-9a-f]{64}$": {
+              "type": "object",
+              "required": ["keyid_hash_algorithms", "keytype", "keyval", "scheme"],
+              "additionalProperties": false,
+              "properties": {
+                "keyid_hash_algorithms": {
+                  "type": "array",
+                  "items": {
+                    "enum": [
+                      "sha256",
+                      "sha512"
+                    ]
+                  }
+                },
+                "keytype": {
+                  "enum": ["rsa"]
+                },
+                "keyval": {
+                  "type": "object",
+                  "required": ["public"],
+                  "additionalProperties": false,
+                  "properties": {
+                    "public": {
+                      "type": "string",
+                      "pattern": "^-----BEGIN PUBLIC KEY-----"
+                    }
+                  }
+                },
+                "scheme": {
+                  "enum": ["rsassa-pss-sha256"]
+                }
+              }
+            }
+          }
+        },
+        "roles": {
+          "type": "object",
+          "required": ["root", "snapshot", "targets", "timestamp"],
+          "additionalProperties": false,
+          "properties": {
+            "root": {
+              "type": "object",
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 64,
+                    "maxLength": 64,
+                    "pattern": "^[a-f0-9]{64}$"
+                  }
+                },
+                "threshold": {
+                  "type": "number",
+                  "minimum": 1
+                }
+              }
+            },
+            "snapshot": {
+              "type": "object",
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 64,
+                    "maxLength": 64,
+                    "pattern": "^[a-f0-9]{64}$"
+                  }
+                },
+                "threshold": {
+                  "type": "number",
+                  "minimum": 1
+                }
+              }
+            },
+            "targets": {
+              "type": "object",
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 64,
+                    "maxLength": 64,
+                    "pattern": "^[a-f0-9]{64}$"
+                  }
+                },
+                "threshold": {
+                  "type": "number",
+                  "minimum": 1
+                }
+              }
+            },
+            "timestamp": {
+              "type": "object",
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 64,
+                    "maxLength": 64,
+                    "pattern": "^[a-f0-9]{64}$"
+                  }
+                },
+                "threshold": {
+                  "type": "number",
+                  "minimum": 1
+                }
+              }
+            }
+          }
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]$"
+        },
+        "version": {
+          "type": "number",
+          "minimum": 0
+        }
+      }
+    }
+  }
+}
--- a/schemas/snapshot.schema.json
+++ b/schemas/snapshot.schema.json
@@ -1,0 +1,704 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "signatures": {
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "keyid": {
+              "type": "string",
+              "minLength": 64,
+              "maxLength": 64,
+              "pattern": "^[a-f0-9]{64}$"
+            },
+            "sig": {
+              "type": "string",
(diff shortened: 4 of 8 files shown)


schemas/targets.schema.json
```diff
@@ -0,0 +1,274 @@
...
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]$"
+        },
+        "targets": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^in-toto-metadata/[0-9a-z.]+.layout$": {
```
trishankatdatadog, line 155: Generally, neither specific targets nor delegations are expected in `targets.json`.

trishankatdatadog (changes requested):
This is a great start, thanks, Fridolin, will really help the community!

I have two comments in general:
1. Some of the assumptions (e.g., about version numbers and keyids) are incorrect with respect to the spec.
2. The other assumptions are not incorrect, but they are specific to the [Datadog Agent integrations](https://www.datadoghq.com/blog/engineering/secure-publication-of-datadog-agent-integrations-with-tuf-and-in-toto/), which will not be useful in general here.

Could we please make these changes and try again? Thanks!

schemas/in-toto-metadata-signer-x.schema.json
```diff
@@ -0,0 +1,145 @@
+{
```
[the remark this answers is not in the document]
fridex: Adressed in 24dc032.

schemas/wheels-signer-x.schema.json
```diff
@@ -0,0 +1,148 @@
+{
```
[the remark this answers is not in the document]
fridex: Adressed in 24dc032.

schemas/in-toto-metadata-signer.schema.json
```diff
@@ -0,0 +1,168 @@
+{
```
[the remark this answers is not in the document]
fridex: Adressed in 24dc032.

schemas/wheels-signer.schema.json
```diff
@@ -0,0 +1,174 @@
+{
```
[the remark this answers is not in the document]
fridex: Adressed in 24dc032.

schemas/timestamp.schema.json
```diff
@@ -0,0 +1,116 @@
+{
```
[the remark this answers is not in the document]
fridex: Made two adjustments to this file based on comments from other files, see 767c54a and 885ea58.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+                  "type": "number",
+                  "minimum": 1
+                }
+              }
+            }
+          }
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]$"
```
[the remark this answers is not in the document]
fridex: Adjusted in 6b6d6fb. Patch versions are now optional and can have more than one digit.

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,704 @@
...
+        },
+        "expires": {
+          "type": "string",
+          "format": "date-time"
+        },
+        "meta": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "in-toto-metadata-signer-0.json": {
```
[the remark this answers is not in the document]
fridex: Adjusted in 850f864.

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,704 @@
...
+              "type": "object",
+              "additionalProperties": false,
+              "properties": {
+                "version": {
+                  "type": "integer",
+                  "minimum": 0
```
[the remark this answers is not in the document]
fridex: Adjusted in 40247c4.

schemas/targets.schema.json
fridex: Adjusted in 975c34a.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+      "properties": {
+        "_type": {
+          "enum": ["root"]
+        },
+        "consistent_snapshot": {
+          "type": "boolean"
+        },
+        "expires": {
+          "type": "string",
+          "format": "date-time"
```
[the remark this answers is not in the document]
fridex: `date-time` should follow [ISO8601](https://www.iso.org/iso-8601-date-and-time-format.html) as in the TUF spec. Checking the TUF spec, `Z` suffix is required. Adjusted to explicitly have this pattern (might lose some datetime semantics though). See 4c2225629.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+                  "type": "array",
+                  "items": {
+                    "enum": [
+                      "sha256",
+                      "sha512"
+                    ]
+                  }
+                },
+                "keytype": {
+                  "enum": ["rsa"]
```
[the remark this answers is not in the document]
fridex: Adjusted in fb56026.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 64,
+                    "maxLength": 64,
+                    "pattern": "^[a-f0-9]{64}$"
```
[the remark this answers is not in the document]
fridex: Adjusted in fb56026.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+            }
+          }
+        },
+        "spec_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]$"
+        },
+        "version": {
+          "type": "number",
+          "minimum": 0
```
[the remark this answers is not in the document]
fridex: Adjusted in [40247c4](https://github.com/theupdateframework/specification/pull/246/commits/40247c49fbcdb0db3d002c909654db3e2aa080df).

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+                  "enum": ["rsa"]
+                },
+                "keyval": {
+                  "type": "object",
+                  "required": ["public"],
+                  "additionalProperties": false,
+                  "properties": {
+                    "public": {
+                      "type": "string",
+                      "pattern": "^-----BEGIN PUBLIC KEY-----"
```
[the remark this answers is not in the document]
fridex: Adjusted in 3a253ff.

schemas/root.schema.json
```diff
@@ -0,0 +1,181 @@
...
+                    }
+                  }
+                },
+                "scheme": {
+                  "enum": ["rsassa-pss-sha256"]
```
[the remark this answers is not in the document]
fridex: Adjusted in 3a253ff. It looks like the standard is not strict - it does not explicitly list supported schemes.

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,704 @@
...
+          "type": "string",
+          "format": "date-time"
+        },
+        "meta": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "in-toto-metadata-signer-0.json": {
+              "type": "object",
+              "additionalProperties": false,
```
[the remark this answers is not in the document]
fridex: Added missing parts in 84ad9f4.

fridex:
Thanks for the review, Trishank. Comments raised were addressed.

Please see also additional changes outside of the PR review comments (starting from d6731dc).

A concern I have is the listing of hashes in `.signed.meta.snapshot\.json` in `snapshots.json`. The schema proposed requires both sha256 and sha512, but the example from the specification lists only sha256.

Also, for completeness, should we include a schema for mirrors.json?

trishankatdatadog:
> Also, for completeness, should we include a schema for mirrors.json?

Sorry, no need for mirrors, which is being replaced by [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). A schema for the latter would be nice!

schemas/root.schema.json
```diff
@@ -0,0 +1,192 @@
...
+        "keys": {
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[0-9a-f]{64}$": {
+              "type": "object",
+              "required": ["keyid_hash_algorithms", "keytype", "keyval", "scheme"],
+              "additionalProperties": false,
+              "properties": {
+                "keyid_hash_algorithms": {
```
trishankatdatadog, line 53: Let's get rid of this, please! It is not in the spec, and is the source of [much confusion](https://github.com/theupdateframework/go-tuf/pull/369).

schemas/targets.schema.json
```diff
@@ -0,0 +1,157 @@
...
+  "additionalProperties": false,
+  "properties": {
+    "signatures": {
+      "type": "array",
+      "items": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "keyid": {
```
trishankatdatadog, line 13: This is a valid assumption [for now](https://theupdateframework.github.io/specification/v1.0.30/index.html#role-keyid), but note that it will change in the [future](https://github.com/theupdateframework/taps/blob/master/tap12.md).

trishankatdatadog:
Thanks, @fridex! If not too much to ask, do you think we could do some sort of integration testing against the Datadog Agent Integrations and/or Sigstore TUF metadata?

Cc @asraa @patflynn

schemas/root.schema.json
fridex: Removed in a5f6a0e. Thanks.

fridex:
> Thanks, @fridex! If not too much to ask, do you think we could do some sort of integration testing against the Datadog Agent Integrations and/or Sigstore TUF metadata?

These schemas pass for Datadog Agent Integrations. I'm open to do the same for Sigstore TUF metadata if you point me to them.

fridex:
> > Also, for completeness, should we include a schema for mirrors.json?
> 
> Sorry, no need for mirrors, which is being replaced by [TAP 4](https://github.com/theupdateframework/taps/blob/master/tap4.md). A schema for the latter would be nice!

Added also schema for TAP 4. To my understanding, the file is located on client machines and thus does not need to be signed. Hence, it keeps just entries from the example in TAP 4 spec.

trishankatdatadog:
> These schemas pass for Datadog Agent Integrations. I'm open to do the same for Sigstore TUF metadata if you point me to them.

Great! @patflynn do you know where we can find the Sigstore metadata?

trishankatdatadog:
@fridex oh, one more thing: could we pls link to these schemas from the spec itself?

schemas/root.schema.json
```diff
@@ -0,0 +1,183 @@
...
+            "type": "string",
+            "minLength": 64,
+            "maxLength": 64,
+            "pattern": "^[a-f0-9]{64}$"
+          },
+          "sig": {
+            "type": "string",
+            "minLength": 1024,
+            "maxLength": 1024,
+            "pattern": "^[a-f0-9]{1024}$"
```
fridex, line 24: One more thing to clarify - do we want to relax requirements on the signature? The spec does not talk about its length.

trishankatdatadog: Yes: should just be a positive length

schemas/root.schema.json
```diff
@@ -0,0 +1,183 @@
...
+          "type": "object",
+          "additionalProperties": false,
+          "patternProperties": {
+            "^[0-9a-f]{64}$": {
+              "type": "object",
+              "required": ["keytype", "keyval", "scheme"],
+              "additionalProperties": true,
+              "properties": {
+                "keytype": {
+                  "enum": ["rsa"]
```
trishankatdatadog, line 54: While there are no mandates, please look [here](https://theupdateframework.github.io/specification/v1.0.30/index.html#keytype) for commonly used values

schemas/root.schema.json
```diff
@@ -0,0 +1,183 @@
...
+                  "required": ["public"],
+                  "additionalProperties": false,
+                  "properties": {
+                    "public": {
+                      "type": "string",
+                      "minLength": 1
+                    }
+                  }
+                },
+                "scheme": {
```
trishankatdatadog, line 67: Please see [here](https://theupdateframework.github.io/specification/v1.0.30/index.html#scheme)

schemas/root.schema.json
```diff
@@ -0,0 +1,183 @@
...
+            "mirror": {
+              "type": "object",
+              "required": ["keyids", "threshold"],
+              "additionalProperties": false,
+              "properties": {
+                "keyids": {
+                  "type": "array",
+                  "items": {
+                    "type": "string",
+                    "minLength": 1
```
trishankatdatadog, line 89: I think we can safely reuse the same keyid requirement as above (SHA2-256)

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,95 @@
...
+        },
+        "expires": {
+          "type": "string",
+          "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$"
+        },
+        "meta": {
+          "type": "object",
+          "additionalProperties": true,
+          "properties": {
+            "targets.json": {
```
trishankatdatadog, line 48: While the spec doesn't mandate a specific file format (e.g., JSON) or filename (e.g., "targets.json"), I think this is acceptable for now.

schemas/timestamp.schema.json
```diff
@@ -0,0 +1,116 @@
...
+                      "pattern": "^[a-f0-9]{64}$"
+                    },
+                    "sha512": {
+                      "type": "string",
+                      "minLength": 128,
+                      "maxLength": 128,
+                      "pattern": "^[a-f0-9]{128}$"
+                    }
+                  },
+                  "required": [
```
trishankatdatadog, line 69: These are not strictly required. Typically, we use these hashes, but they are not required.

schemas/targets.schema.json
```diff
@@ -0,0 +1,157 @@
...
+                      "pattern": "^[a-f0-9]{64}$"
+                    },
+                    "sha512": {
+                      "type": "string",
+                      "minLength": 128,
+                      "maxLength": 128,
+                      "pattern": "^[a-f0-9]{128}$"
+                    }
+                  },
+                  "required": [
```
trishankatdatadog, line 121: Again, not required.

schemas/targets.schema.json
```diff
@@ -0,0 +1,157 @@
...
+                    "sha512"
+                  ]
+                },
+                "length": {
+                  "type": "integer",
+                  "minimum": 1
+                }
+              },
+              "required": [
+                "custom",
```
trishankatdatadog, line 132: not required

fridex: Fixed in c423ee29196cb2ffd5c0163378276a7adfef3333

schemas/targets.schema.json
fridex: Fixed in 39d2addfba0643f0ff719dc9b753d33ac671c6f4.

schemas/timestamp.schema.json
fridex: Fixed in 39d2addfba0643f0ff719dc9b753d33ac671c6f4.

schemas/root.schema.json
fridex: Adjusted in 1e6a12972a038a7d034012430a3010377a585ef8.

schemas/root.schema.json
fridex: Adjusted in 1e6a12972a038a7d034012430a3010377a585ef8.

schemas/root.schema.json
fridex: Adjusted in 1e6a12972a038a7d034012430a3010377a585ef8.

fridex:
> Sigstore's current TUF repository is here: https://github.com/sigstore/root-signing/tree/main/repository/repository!

The schema validation passes except for `date-time` string in `expires`:

```
jsonschema.exceptions.ValidationError: '2021-12-18T13:28:12.99008-06:00' does not match '^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$'

Failed validating 'pattern' in schema['properties']['signed']['properties']['expires']:
    {'pattern': '^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$',
     'type': 'string'}

On instance['signed']['expires']:
    '2021-12-18T13:28:12.99008-06:00'
```

The spec says in the [General principles section](https://theupdateframework.github.io/specification/v1.0.30/index.html#file-formats-general-principles):

> Metadata date-time follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example [date-time](https://theupdateframework.github.io/specification/v1.0.30/index.html#date-time) string is "1985-10-21T01:21:00Z".

I've put back `date-time` format that covers also cases that are used in the root-signing repository in cc2c17. If the date-time paragraph in the spec is applicable here, my guess is this needs a clarification.

trishankatdatadog:
> > Sigstore's current TUF repository is here: https://github.com/sigstore/root-signing/tree/main/repository/repository!
> 
> The schema validation passes except for `date-time` string in `expires`:
> 
> ```
> jsonschema.exceptions.ValidationError: '2021-12-18T13:28:12.99008-06:00' does not match '^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$'
> 
> Failed validating 'pattern' in schema['properties']['signed']['properties']['expires']:
>     {'pattern': '^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$',
>      'type': 'string'}
> 
> On instance['signed']['expires']:
>     '2021-12-18T13:28:12.99008-06:00'
> ```
> 
> The spec says in the [General principles section](https://theupdateframework.github.io/specification/v1.0.30/index.html#file-formats-general-principles):
> 
> > Metadata date-time follows the ISO 8601 standard. The expected format of the combined date and time string is "YYYY-MM-DDTHH:MM:SSZ". Time is always in UTC, and the "Z" time zone designator is attached to indicate a zero UTC offset. An example [date-time](https://theupdateframework.github.io/specification/v1.0.30/index.html#date-time) string is "1985-10-21T01:21:00Z".
> 
> I've put back `date-time` format that covers also cases that are used in the root-signing repository in [bcc2c17](https://github.com/theupdateframework/specification/commit/bcc2c170c56a3f2444e8897443909b06e0365f58). If the date-time paragraph in the spec is applicable here, my guess is this needs a clarification (in TUF spec/sigstore/root-signing repo/TUF implementation).

I'd say the original sigstore metadata was incorrect in not following the TUF spec. Best to use the `Z` suffix specified in the spec. WDYT @asraa ?

trishankatdatadog (approved):
Well done, Fridolin, I can't think of any more complaints! 🎉

fridex:
> Well done, Fridolin, I can't think of any more complaints! 🎉

Thanks for your patience and thorough review, Trishank 👍🏻

asraa:
> I'd say the original sigstore metadata was incorrect in not following the TUF spec. Best to use the Z suffix specified in the spec. WDYT @asraa ?

And yes! We only got expiraiton TUF compliant in later roots, so early roots should not be compliant.

trishankatdatadog:
@patflynn would you be interested in reviewing this PR? 🙂

patflynn:
Hi there! Sorry I missed all the pings on this thread! I'm really grateful for @fridex's contribution here. As long as the current historical set of metadata resources parse correctly I think we're doing as well as we could expect!  LGTM.

For some context we discussed at Kubecon potentially moving sigstore TUF schema definition to proto at a future date, but I don't think that should block this.

trishankatdatadog:
Can we please get an amen from @lukpueh, @joshuagl, or @mnm678? 🙏🏽

trishankatdatadog:
I think somewhere we should also document that editors should update these JSON schema files whenever they update the metadata formats in the spec. Ideally, there should be code in GitHub Workflows that checks for this, but documentation will do fornow.

trishankatdatadog:
@fridex could we please resolve conflicts in the meantime?

lukpueh:
Great work, @fridex! What do you think about defining subschemas for re-occuring parts? The TUF spec has many of those and JSON schema seems to [support it](https://json-schema.org/understanding-json-schema/structuring.html). It would make review and maintenance a lot easier.

jku:
I have done a few attempts at reviewing (and I can post some detail questions if you'd like them before I finish) but I feel like a _"LGTM"_ without seeing results from a public validation test suite or real commitment from an open source project to use this data as part of a test suite, seems like a big ask from reviewers.

Basically, is there a plan/commitment to use these schemas somewhere? If not, is there value in adding 600 lines of hard to validate schema -- I think it's unlikely to match the implementations without continuous testing. 

Would it make more sense to add a test suite (wherever it's planned) first, and then propose that the spec repo should maintain the schemas?

Obviously, I'm not a spec maintainer so if they are ok with this then my reservations are moot...

trishankatdatadog:
> I have done a few attempts at reviewing (and I can post some detail questions if you'd like them before I finish) but I feel like a _"LGTM"_ without seeing results from a public validation test suite or real commitment from an open source project to use this data as part of a test suite, seems like a big ask from reviewers.
> 
> Basically, is there a plan/commitment to use these schemas somewhere? If not, is there value in adding 600 lines of hard to validate schema -- I think it's unlikely to match the implementations without continuous testing.
> 
> Would it make more sense to add a test suite (wherever it's planned) first, and then propose that the spec repo should maintain the schemas?
> 
> Obviously, I'm not a spec maintainer so if they are ok with this then my reservations are moot...

These are good reservations, and we have thought about them. We do have an internal app that uses and tests against these schemas. A public test suite would be nice, but demands a fairly representative dataset. @fridex WDYT?

lukpueh:
IMO it makes sense to not tie schemas hosted in the repo to schemas used in a test suite of an implementation. As @jku says, they might not even match. But wouldn't it be a good thing if there were reference schemas that made such a deviation apparent?

I agree that they are prone to become outdated if we don't automatically check spec changes against the schemas, which is hard to do because the format definitions don't use a standard language. Btw. same is true for the example metadata in the spec. Not sure how to solve this problem, without changing the format definition language. Maybe we could at least run the schemas against the example metadata?

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,93 @@
...
```
jku, line 46: I would expect you could match all items in the object with patternproperties, not just the "targets.json" field ?

schemas/snapshot.schema.json
```diff
@@ -0,0 +1,93 @@
...
+                "version": {
+                  "type": "integer",
+                  "minimum": 1
+                },
+                "length": {
+                  "type": "integer",
+                  "minimum": 1
+                },
+                "hashes": {
+                  "type": "object"
```
jku, line 59: This is defined in targets schema but not here

schemas/targets.schema.json
```diff
@@ -0,0 +1,178 @@
...
+                  },
+                  "paths": {
+                    "type": "array",
+                    "minLength": 1
+                  },
+                  "terminating": {
+                    "type": "boolean"
+                  }
+                },
+                "required": ["name", "keyids", "threshold", "paths", "terminating"]
```
jku, line 106: paths is not required

jku (commented):
After reading through the whole thing I'm more convinced that the schema needs to be tested against metadata from multiple actual implementations. If it's not, the schema will contain bugs: they are really hard for humans to spot in this format. 

I don't mean that there needs to be an automated test suite and I don't mean the tests should pass (disagreements about correctness are quite likely)... but some reviewed results from using the schema to validate metadata seem necessary to have trust on the schema.

lukpueh:
>  We could do something like runnable docstrings: we could decorate the example metadata written in the spec in such a way that when you compile the spec into a readable format, tests would run the example metadata against the schemas. Do I make sense?

You do, @trishankatdatadog! Here's how I did something similar for an in-toto demo, that is [running a script in CI](https://github.com/in-toto/demo/blob/adf5c5621523dbc9f88dbf23e6cffb8683c0dcf5/.github/workflows/ci.yml#L41) that [regex matches snippets from a markdown doc and asserts for an expected output](https://github.com/in-toto/demo/blob/adf5c5621523dbc9f88dbf23e6cffb8683c0dcf5/run_demo_md.py#L32-L36).

lukpueh:
So we have two suggestions here:

1. Run the schema against real metadata from different implementations at least once and reference those static results e.g. in this PR, so that we can be more certain about the correctness of the schemas.
2. Setup a CI job to "run schemas against the spec", to avoid that schemas become outdated when the spec changes. 
    -> This is less trivial, because we can't actually just run schemas against the spec. We could run them against example metadata as described above, but then we'd still have to rely on spec changes being synced with either examples or schemas to detect things going out of sync.
 
Maybe we should start with 1. and  ticketize 2.

joshuagl:
I am all for including the schemas in this organisation as a starting point for implementers (the context in which this came up, iirc). I would like to see them updated to use subschemas for reoccurring parts first as @lukpueh suggested and it is important to see them validated against real data as @jku suggested.

Given that we will not, at least to start, have a mechanism to test them over time I think it's worth putting them in a subdirectory which indicates they are reference material/a starting point such as `contrib`. We could even consider including them in the reference implementation (python-tuf) and adding a test that the schema remain valid against metadata files generated by python-tuf?

Thanks for working on this @fridex!

trishankatdatadog:
> Given that we will not, at least to start, have a mechanism to test them over time I think it's worth putting them in a subdirectory which indicates they are reference material/a starting point such as `contrib`. We could even consider including them in the reference implementation (python-tuf) and adding a test that the schema remain valid against metadata files generated by python-tuf?

I think it's reasonable to start with this, and include `specification` as a submodule you could use to run schema tests against `python-tuf`. Then, we file issues to include more real-world tests. WDYT?

fridex:
Thanks all for review comments. I'll incorporate suggestions.

> > Given that we will not, at least to start, have a mechanism to test them over time I think it's worth putting them in a subdirectory which indicates they are reference material/a starting point such as `contrib`. We could even consider including them in the reference implementation (python-tuf) and adding a test that the schema remain valid against metadata files generated by python-tuf?
> 
> I think it's reasonable to start with this, and include `specification` as a submodule you could use to run schema tests against `python-tuf`. Then, we file issues to include more real-world tests, split schema into reusable subschemas, and so on. WDYT?

IMHO it is very reasonable to test JSON schema files as discussed. If python-tuf is the referenced implementation, it might be a good idea to have tests with JSON schema included there (and later let other implementations use it as well). I'm personally not a big fan of Git submodules, however if we could create an automation (a GitHub action maybe?) that would keep the submodule up-to-date, that would be great. Semantically, the specification repo could be the right place for schema files. WDYT?

lukpueh:
> IMHO it is very reasonable to test JSON schema files as discussed. If python-tuf is the referenced implementation, it might be a good idea to have tests with JSON schema included there (and later let other implementations use it as well). I'm personally not a big fan of Git submodules, however if we could create an automation (a GitHub action maybe?) that would keep the submodule up-to-date, that would be great. Semantically, the specification repo could be the right place for schema files. WDYT?

Let's take one step at a time:

1. DRY schemas (subschemas)
2. Run schemas against an exemplary set of metadata (e.g. from spec example metadata, python-tuf, or go-tuf) for initial anecdotal validation
3. Auto-run schemas against spec metadata (see "regex snippets" suggestions above)
4. Auto-run schemas against implementation metadata  (see submodule suggestions above)

IMO 1 and 2 are blockers for this PR and 3 and 4 is future work.

> 
> EDIT: Are signed commits required in the repo?

Does not look like it :)

PR #281, opened 2023-05-15
Fix specification builds

joshuagl:
Something changed in our GitHub Actions environment (either the platform itself or our organisation's default permissions) since the last specification release such that the build and publish workflow has failed for the most recent change.

This PR is the simplest fix, adding the required permissions for the workflow to succeed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ jobs:
     name: Make and publish spec release
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: write # required to publish (tag) a release 
     steps:
     - name: Clone main
       uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab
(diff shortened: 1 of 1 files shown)


joshuagl:
Hi @lukpueh, @mnm678, @trishankatdatadog -- I'd appreciate a quick review on this one. Our last set of changes are not reflected in the published spec because the GHA doesn't work without this permission being granted.

.github/workflows/release.yml
```diff
@@ -8,6 +8,8 @@ jobs:
   make-release:
     name: Make and publish spec release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
joshuagl, line 12:
```suggestion
      contents: write # required to publish (tag) a release 
```

lukpueh (dismissed):

.github/workflows/release.yml
lukpueh: Good catch! Did GitHub not inform us that our last builds failed?

joshuagl: our most recent builds exited early because the spec itself had not changed (dependency updates and affiliation update).

lukpueh (approved):

PR #284, opened 2023-06-26
Proposal: implementation roadmaps

mnm678:
This is a proposal for organizing a place to collect roadmaps from different TUF implementations to provide discoverability of these projects and their plans.

Once this idea (and the location of this list) is discussed with the community, I will follow up with projects to collect an initial list of roadmaps.

### DIFF
--- a/Roadmap.md
+++ b/Roadmap.md
@@ -35,3 +35,11 @@ Which of these should block releases?
 * TAP 3 & 4, work on implementations
 * Implement TAP 13 in go-tuf
 * All implementations should catch up to 1.x.y by $MONTH and $DATE?
+
+## Implementation roadmaps
+
+The following are links to roadmaps for various TUF implementations. These roadmaps are maintained by the projects. If you have a TUF implementation, feel free to open a pr to add it here.
+
+* **python-tuf**: 
+* **go-tuf**: https://github.com/orgs/theupdateframework/projects/3
+* ...


Roadmap.md
```diff
@@ -35,3 +35,11 @@ Which of these should block releases?
 * TAP 3 & 4, work on implementations
 * Implement TAP 13 in go-tuf
 * All implementations should catch up to 1.x.y by $MONTH and $DATE?
+
+## Implementation roadmaps
+
+The following are links to roadmaps for various TUF implementations. These roadmaps are maintained by the projects. If you have a TUF implementation, feel free to open a pr to add it here.
+
+* **python-tuf**: 
+* **go-tuf**: https://github.com/orgs/theupdateframework/projects/3
```
rdimitrov, line 44: I know we don't have a dedicated roadmap yet, but it's worth noting - should we assume that a roadmap (something more high-level) would be the same as a project plan (i.e. like what we have so far in go-tuf which is a bit more detailed)?

trishankatdatadog (approved):

Roadmap.md
znewman01: I would like go-tuf to have a higher-level roadmap. I think that would serve a different-but-important use compared with the project board.

znewman01 (commented):
I think this is a great idea. Once question: is there a process by which we can make sure these don't get _too_ stale? (Including the TUF-wide one above.) I think "review the roadmap every month at the meeting" doesn't work. Maybe an automatic issue every 6 months or so?

Roadmap.md
rdimitrov: +1 for that 👍 I imagine the roadmap as a high-level plan that is being revisited every quarter or so.

lukpueh:
> Once this idea (and the location of this list) is discussed with the community, [...]

IMO  `github.com/theupdateframework/community` would be a good location. It does not exist yet, but I just opened a [poll on creating/naming such a repo](https://cloud-native.slack.com/archives/C8NMD3QJ3/p1688028555729659).


> Maybe an automatic issue every 6 months or so?

related example: https://github.com/theupdateframework/python-tuf/blob/develop/.github/workflows/maintainer-permissions-reminder.yml

rdimitrov (approved):

mnm678:
I created #theupdateframework/community#4 to supersede this issue.

PR #296, opened 2023-12-06
Update release.yml

trishankatdatadog:
Defer setting up Python (and, more importantly, failing to clean up a nonexistent pip cache dir when no Python packages have been installed) if there is no new spec to release.

Otherwise, you get spurious errors like [this](https://github.com/theupdateframework/specification/actions/runs/7116550212/job/19375232136).


trishankatdatadog:
I _believe_ it works...

PR #301, opened 2024-02-23
Update wildcard pattern behavior

adityasaky:
This PR adds explicit examples of targets that are not matched by a wildcard. It also adds a note warning users that incorrect assumptions about wildcard behavior can potentially lead to an untrusted role signing for a target.

Related: https://github.com/theupdateframework/go-tuf/issues/590

### DIFF
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -3,7 +3,7 @@ Title: The Update Framework Specification
 Shortname: TUF
 Status: LS
 Abstract: A framework for securing software update systems.
-Date: 2023-04-14
+Date: 2024-02-23
 Editor: Justin Cappos, NYU
 Editor: Trishank Karthik Kuppusamy, Datadog
 Editor: Joshua Lock, Verizon
@@ -1093,8 +1093,17 @@ attributes are OPTIONAL, if used, exactly one of them should be set.
       `"foo-version-2.tgz"` and `"foo-version-a.tgz"`, but not
       `"foo-version-alpha.tgz"`.
     * a <a>PATHPATTERN</a> of `"*.tgz"` would match `"foo.tgz"` and `"bar.tgz"`,
-      but not `"targets/foo.tgz"`
-    * a <a>PATHPATTERN</a> of `"foo.tgz"` would match only `"foo.tgz"`
+      but not `"targets/foo.tgz"`.
+    * a <a>PATHPATTERN</a> of `"foo.tgz"` would match only `"foo.tgz"`.
+    * a <a>PATHPATTERN</a> of `"foo/*"` matches `"foo/bar.tgz"` but not
+      `"foo/baz/bar.tgz"`, `"foo/bar/baz/bar.tgz"`, and so on.
+
+    Note: It is important to understand the functioning of path patterns to
+    avoid mistaken assumptions that can lead to a false sense of security. For
+    example, an assumption that `"foo/*"` applies recursively to all files in
+    subdirectories of `foo` in a terminating delegation could allow a subsequent
+    delegated role that should not be trusted to sign for a target.
+
 
 
 Prioritized delegations allow clients to resolve conflicts between delegated
--- a/tuf-spec.md
+++ b/tuf-spec.md
@@ -16,7 +16,7 @@ Boilerplate: copyright no, conformance no
 Local Boilerplate: header yes
 Markup Shorthands: css no, markdown yes
 Metadata Include: This version off, Abstract off
-Text Macro: VERSION 1.0.33
+Text Macro: VERSION 1.0.34
 </pre>
 
 Note: We strive to make the specification easy to implement, so if you come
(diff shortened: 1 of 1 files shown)


tuf-spec.md
```diff
@@ -1093,8 +1093,17 @@ attributes are OPTIONAL, if used, exactly one of them should be set.
...
       `"foo-version-alpha.tgz"`.
     * a <a>PATHPATTERN</a> of `"*.tgz"` would match `"foo.tgz"` and `"bar.tgz"`,
-      but not `"targets/foo.tgz"`
-    * a <a>PATHPATTERN</a> of `"foo.tgz"` would match only `"foo.tgz"`
+      but not `"targets/foo.tgz"`.
+    * a <a>PATHPATTERN</a> of `"foo.tgz"` would match only `"foo.tgz"`.
+    * a <a>PATHPATTERN</a> of `"foo/*"` matches `"foo/bar.tgz"` but not
+      `"foo/baz/bar.tgz"`, `"foo/bar/baz/bar.tgz"`, and so on.
+
+    Note: It is important to understand the functioning of path patterns to
```
trishankatdatadog, line 1101: I'd also list a note to the opposite effect _first_: remind the reader _why_ this behaviour is there (not for arbitrary reasons).

trishankatdatadog (changes requested):

tuf-spec.md
adityasaky: I tweaked the text a bit, does it help?

mnm678 (dismissed):
waiting on the workflow run, but lgtm

adityasaky:
i bumped the patch version number.

tuf-spec.md
trishankatdatadog: No, sorry. First, please briefly explain why the current patterns work the way they do, and as such, they are optimised for those anticipated use cases. Your use cases here are outside of that, and should thus fall under a second paragraph.

adityasaky: I'm not the best person to discuss why the current patterns work the way they do as I don't actually have that context. @mnm678 can you help with some text here? Thanks!

tuf-spec.md
```diff
@@ -1093,8 +1093,19 @@ attributes are OPTIONAL, if used, exactly one of them should be set.
...
+    * a <a>PATHPATTERN</a> of `"foo/*"` matches `"foo/bar.tgz"` but not
+      `"foo/baz/bar.tgz"`, `"foo/bar/baz/bar.tgz"`, and so on.
+
+    Note: As each <PATHPATTERN> describes the paths that a delegated role is
+    trusted to provide, it is important to understand the functioning of path
+    patterns to avoid mistaken assumptions that can lead to a false sense of
+    security. For example, an assumption that `"foo/*"` applies recursively to
+    all files in subdirectories of `foo` in a terminating delegation could allow
+    a subsequent delegated role that should not be trusted to sign for a target
+    in a subdirectory of `foo`.
```
joshuagl, line 1107: This is a great clarification to add, thanks! This point on wildcards not applying recursively feels important enough that I wonder whether we should raise it above the examples so that it's a little more prominent, WDYT?

joshuagl (dismissed):
This is great, thank you. Approved as is, but I wonder if we should move the coment on wildcards not apply recursively above the examples? 

It is somewhat explicit because we state that we follow shell style globbing, but for many shell is synonymous with bash, which has [globstar](https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html).

tuf-spec.md
joshuagl, line 1087: Adding a suggestion to move the Note before the examples as previously suggested.

```suggestion
    NOT be matched by a wildcard in the <a>PATHPATTERN</a>.

    Note: As each <PATHPATTERN> describes the paths that a delegated role is
    trusted to provide, it is important to understand the functioning of path
    patterns to avoid mistaken assumptions that can lead to a false sense of
    security. For example, an assumption that `"foo/*"` applies recursively to
    all files in subdirectories of `foo` in a terminating delegation could allow
    a subsequent delegated role that should not be trusted to sign for a target
    in a subdirectory of `foo`.
```

tuf-spec.md
```diff
@@ -1097,8 +1097,19 @@ attributes are OPTIONAL, if used, exactly one of them should be set.
...
```
joshuagl:
```suggestion
```
As suggested, move the Note to before the examples. This could help people who skim examples.

joshuagl (commented):
This change looks good and received an LGTM from two maintainers, so I took the liberty of making suggestions to address the only remaining review comment in the hope we can land this.

tuf-spec.md
```diff
@@ -1090,15 +1090,27 @@ attributes are OPTIONAL, if used, exactly one of them should be set.
     also recommended for <a>TARGETPATH</a>. A path separator in a path SHOULD
     NOT be matched by a wildcard in the <a>PATHPATTERN</a>.
 
+    Note: As each <PATHPATTERN> describes the paths that a delegated role is
```
joshuagl: Suggesting a fix for an incorrect reference to a definition
```suggestion
    Note: As each <a>PATHPATTERN</a> describes the paths that a delegated role is
```

bigdaddyliz101-sys:
Credit.txt

PR #316, opened 2026-03-27
spec: s6.1: recommend aggressive timestamp key rotation

cyphar:
The previous text tried to minimise the risk of the root.json freeze
attack, but in practice the expiry of root.json is set to several years
to reduce the usage of the incredibly critical root role keys -- if an
attacker freezes root.json they can buy themselves time to try to access
a threshold of keys even if you try to key rotate your way out of it.

The core issue that allows the root.json freeze attack is that the new
root.json is the only thing signed by the new root keys, so there is no
way for a user to detect that they are being blocked from seeing a new
root.json.

However, a very simple workaround (and arguably good general policy) is
to rotate other keys that are used to sign (comparatively) short-expiry
metadata. By rotating all of the timestamp keys, an attacker blocking
root.json will cause clients to error out as soon as they either see a
new timestamp.json (because it is signed by keys the client does not
recognise) or an old timestamp.json after a short delay (because the
short expiry has elapsed). The client cannot conclusively determine that
the attack was actually a root.json freeze attack or just a malicious
mirror providing completely invalid signatures, but the attack is no
longer silent.

Unlike other alterative workarounds (such as including the hash and
revision number of root.json in snapshot.json), this approach does not
open the door to lower-trust keys being able to cause DoSes or
fast-forward attacks by including bogus information about root.json.
This new recommendation just boils down to simply rotating a few extra
keys when updating root.json.

This does open the door to possible races with naive repository storage
media (such as a client erroring out after seeing a mismatched root.json
and timestamp.json, because one file was updated before another) but
those are already the case with any other key rotation and there are
well-known workarounds (such as using transactions to ensure that all
metadata files are updated together).

Signed-off-by: Aleksa Sarai <aleksa@amutable.com>


JustinCappos:
My sense is that a short-lived timestamp role is already understood / done
by adopters.  Where does this recommendation come from?  Ddi you set this
up in your environment and run into this?

cyphar:
> Where does this recommendation come from? Did you set this up in your environment and run into this?

While evaluating TUF for our update scheme at @amutable-systems, this particular attack came to mind and then I re-read the spec and found the operational note for this somewhat underwhelming. For our deployment we plan to use the method I've outlined in this PR, but given that it is a fairly minimal operational change and avoids an attack that the spec currently doesn't really give much guidance about, it seemed reasonable to improve the spec wording and provide this kind of recommendation.

But as discussed on Slack, we can also just put this in a "best practices" doc -- though given that the spec already has this operational section, it makes more sense to me to just adjust the existing text.

> My sense is that a short-lived timestamp role is already understood / done by adopters.

Based on what I've seen, that doesn't really seem to be the case (and you could argue the note in s6.1 of the specification kind of implies that there isn't a different workaround, as it effectively says that expiries are TUF's defence against this attack).

I took a look at the reference implementations (and some non-reference ones like [the `tough` rust crate](https://github.com/awslabs/tough)) and none of them appear to have any kind of logic to do this _implicitly_ so an operator would need to do it manually.

FWIW it seems that Sigstore does this in practice ([there are several revisions with the same timestamp key but they also contain the same keys for every other role](https://github.com/sigstore/root-signing/tree/main/metadata/root_history) and so the root was probably just signed to refresh the expiry) but I didn't see any mention in their documentation as to this being quite an important aspect of the security of key rollovers.

JustinCappos:
I've thought about this a bit and would like this to be a documentation change instead.  My thinking is that one could make the same argument about root not expiring every 10 seconds, etc.

We do already cover this to some extent in FAQ question 8: https://theupdateframework.io/docs/faq/  Is this something you saw?  

If you wanted to work to put together an example setup or similar, we could accept a PR for that for the TUF website.  In this case, please get at least one other adopter to agree with you that the setup you're proposing is sane / standard.

cyphar:
@JustinCappos I did but that part of the FAQ is talking about the metadata itself having short expiries, not rotating the keys themselves. They're two separate problems.

If you don't rotate the keys when changing the root keys, clients are vulnerable to a partial freeze attack (giving an attacker more time to attack your remaining root keys) as long as the old root doesn't expire. This weakens one of the key benefits of TUF's key rotation mechanisms -- even if you detect a root key being exfiltrated, you can't be sure clients will no longer trust that key.

I can ask someone from SigStore to comment on their internal policy for updating their timestamp keys, but as I mentioned above it seems that this is something they already do. :shrug:

JustinCappos:
Okay, would you like to add a FAQ question (or augment what is there)?

Feel free to reach out on slack if you need any help shaping this or getting it in.  I think if it's an FAQ item / clarification, it should be pretty easy.

cyphar:
Sure, I'll write something up.

Issue #319, opened 2026-07-15
Link to Canonical JSON specification is broken


erickt:
I noticed that the link to https://wiki.laptop.org/go/Canonical_JSON in the spec is not working. From archive.org, it seems its been broken since at least March. I did find a snapshot that works at https://web.archive.org/web/20251209150702/http://wiki.laptop.org/go/Canonical_JSON. Hopefully it will come back at some point in time, but in case it doesnt, could we get a copy of OLPC's canonical json in this repository?

joshuagl:
This was resolved in #320, thanks.

PR #320, opened 2026-07-15
Update canonical JSON URL in tuf-spec.md

JustinCappos:
Updated the URL for the canonical JSON reference to an archived version since the main link is broken.

Fix for #319.

~~My sense is that since this isn't changing the text at all, we don't need to bump the version / patch number.~~

We need to update the patch number and date or the linter will complain.


tuf-spec.md
```suggestion
Date: 2026-07-15
```

```suggestion
Text Macro: VERSION 1.0.35
```

joshuagl:
> We need to update the patch number and date or the linter will complain.

We also need to bump them so that the auto-publishing pipeline will push the changes.

joshuagl (approved):

mnm678 (approved):

PR #322, opened 2026-08-05
Only define THRESHOLD once

joshuagl:
#314 is a great clarification, however it also introduced a second definition of THRESHOLD _without_ a <dfn> tag. This definition is duplicative and gets lost because it's not marked up as other definitions.

There are two ways we can fix this:
1. introduce a second dfn tag and use a for attribute on each definition,
  i.e., `<dfn for="root">`
2. streamline to a single definition and update the DELEGATIONS definition
  to reference the THRESHOLD definition in the root section

I've opted for 2 here as it's simpler and avoids redundant duplicate definitions, but please let me know if you would prefer 1.

As well as this, I ended up doing some cleanup in this PR:
- fixed warnings from latest bikeshed
- pinned the version of bikeshed included in our workflows
- updated my affiliation


JustinCappos (approved):
Thanks.  Seems like a good change.  

Usually might ask to break apart this PR, but the other changes are minor enough, so I feel we should just merge.

mnm678 (approved):

